PR #5, opened 2020-08-29
Adds initial Kubebuilder scaffolding

danehans:
Scaffolds the project with Kubebuilder using `projectcontour.io` as the domain, `operator` as the group, `Contour` as the kind and `v1alpha1` as the version.

/assign @jpeach @Miciah


Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Build the manager binary
+FROM golang:1.13 as builder
```
jpeach, line 2: Might as well begin with Go 1.15 :)

jpeach (approved):
This looks good; my only suggestion is that we might as well start off with Go 1.15.

youngnick (approved):
Yeah `operator.projectcontour.io/Contour` LGTM for the apigroup and name.

PR #7, opened 2020-08-31
Upgrade to Go 1.15.

jpeach:
Following on from #5, when starting a new project, we should use the
latest Go release.


skriss:
I guess you'll want to bump the Dockerfile too

PR #10, opened 2020-09-01
Initial Contour Controller Implementation

danehans:
- Moves `main()` to cmd/operator pkg. `cmd/operator/start.go `: Exposes a flag to set the Contour image used by the operator.
- Singularizes the controller pkg.
- `controller/contour/controller.go`: Initial controller implementation that watches the contour CRD and a) finalizes the resource and b) creates the contour namespace if it does not exist.
- `util/retryableerror/retryableerror.go`: Utility function for retrying reconciliation errors.
- `util/slice/slice.go`: Utility function to check if a given slice of strings contains the provided string.

Signed-off-by: Daneyon Hansen <daneyonhansen@gmail.com>

PR #14, opened 2020-09-02
Hoist the controller client into a named field.

jpeach:
Using named fields makes it easier to find where fields are used, which
helps refactoring and code readability.

### DIFF
--- a/controller/contour/controller.go
+++ b/controller/contour/controller.go
@@ -39,7 +39,7 @@ type Config struct {
 // Reconciler reconciles a Contour object.
 type Reconciler struct {
 	Config Config
-	client.Client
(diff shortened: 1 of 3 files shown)


danehans:
@jpeach PR #10 introduced additional client calls. Do you mind updating this PR so it captures all calls? If not, let me know and I'll create another PR that includes the additional calls.

jpeach:
> @jpeach PR #10 introduced additional client calls. Do you mind updating this PR so it captures all calls? If not, let me know and I'll create another PR that includes the additional calls.

This is already on top of #10, so it updates all the call sites.

youngnick (approved):
LGTM

PR #17, opened 2020-09-02
Fixes Operator Build and Deployment Make Targets

danehans:
Previously, the `docker-build` make target would fail due to the `util/` directory not being copied at build time. The PR also fixes the `make deploy` target so the operator can run in-cluster and reconcile `Contour` objects.

/cc @jpeach @Miciah


jpeach (approved):
LGTM

Issue #18, opened 2020-09-02
Improve Namespace Support


danehans:
**Please describe the problem you have**
- As a cluster admin, I need the ability to run multiple instances of Contour in the same namespace.
- As a cluster admin, I need the ability to run Contour in a namespace other than "projectcontour".

See [this](https://kubernetes.slack.com/archives/C8XRH2R4J/p1599072008264500) Slack thread of additional background.

jpeach:
It's possible to deploy multiple contours in the same namespace, or multiple contours in a namespace each. The operator will need to configure leader election carefully to make both these options work. You also need to configure how the data plane choses a contour to connect to.

danehans:
In addition to https://github.com/projectcontour/contour-operator/issues/18#issuecomment-686782485, https://github.com/projectcontour/contour/issues/2122 must be resolved to support multiple contours per namespace.

danehans:
xref: https://kubernetes.slack.com/archives/C8XRH2R4J/p1607972656129200

xpepermint:
Is there a planned milestone for this issue?

danehans:
@xpepermint this issue is blocked by https://github.com/projectcontour/contour/issues/2122 (unprioritized too). If this feature is important to you, can you please describe your use case in 2122?

PR #20, opened 2020-09-03
Adds Namespace Support to Contour API

danehans:
Adds namespace support to the Contour API type. Partially fixes issue https://github.com/projectcontour/contour-operator/issues/18.

/cc @jpeach @Miciah @stevesloka


stevesloka:
@danehans do you need to have namespace be a param or just create the instance of Contour in the same namespace where the CRD is created?

api/v1alpha1/contour_types.go
```diff
@@ -51,6 +51,33 @@ type ContourSpec struct {
 	// +kubebuilder:default=2
 	// +kubebuilder:validation:Minimum=0
 	Replicas int32 `json:"replicas,omitempty"`
+
+	// namespace defines the schema of a Contour namespace.
+	// See each field for additional details.
+	//
+	// +optional
```
jpeach, line 58: Rather than optional, suggest adding a default that sets the namespace to `projectcontour`.

api/v1alpha1/contour_types.go
```diff
@@ -51,6 +51,33 @@ type ContourSpec struct {
 	// +kubebuilder:default=2
 	// +kubebuilder:validation:Minimum=0
 	Replicas int32 `json:"replicas,omitempty"`
+
+	// namespace defines the schema of a Contour namespace.
+	// See each field for additional details.
```
jpeach, line 56: What should happen if multiple Contour resources specify the same namespace?

jpeach:
> @danehans do you need to have namespace be a param or just create the instance of Contour in the same namespace where the CRD is created?

This would remove the possibility of different Contour resources specifying the same namespace. OTOH, it implies a different privilege model for being able to install Contour 🤷

danehans:
> @danehans do you need to have namespace be a param or just create the instance of Contour in the same namespace where the CRD is created?

@stevesloka typically an instance of the CRD, `Contour` in this case, exists in the same namespace as the operator and the operand(s), e.g. the Contour deployment, Envoy daemonset, etc. exist in a separate namespace (projectcontour by default). It's the operand namespace that this PR intends to address.

api/v1alpha1/contour_types.go
danehans: @jpeach I would expect multiple instances of Contour to run in said namespace.

jpeach (approved):
lgtm

PR #21, opened 2020-09-03
Adds Contour RBAC Support

danehans:
Adds support for Contour RBAC:
- Ensures Contour and Envoy service accounts exist.
- Ensures Contour cluster roles exist.
- Ensures cluster roles are bound the service accounts.
- Remove RBAC upon Contour deletion.


jpeach:
> /hold

I'd love to see prow support in the project :)

danehans:
@Miciah PTAL when you have a moment.

controller/contour/rbac.go
```diff
@@ -0,0 +1,263 @@
...
+		Name:      defaultContourServiceAccount,
+	}
+	envoy := types.NamespacedName{
+		Namespace: ns,
+		Name:      defaultEnvoyServiceAccount,
+	}
+	names := []types.NamespacedName{ctr, envoy}
+	for _, name := range names {
+		if err := r.ensureServiceAccount(ctx, name.Namespace, name.Name); err != nil {
+			return fmt.Errorf("failed to create service account %s/%s: %w", name.Namespace, name.Name, err)
```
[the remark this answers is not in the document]
danehans: @Miciah I'm getting ready to push an update that looks like:
```
	for _, name := range names {
		if err := r.ensureServiceAccount(ctx, name.Namespace, name.Name); err != nil {
			return fmt.Errorf("failed to ensure service account for contour %s/%s: %w", contour.Namespace,
				contour.Name, err)
		}
	}
```
This particular error now references the Contour object instead of the ServiceAccount namespace/name. `err` will reference the details of the `ensureServiceAccount()` error. I do like using ns/name since `Contour` and `ServiceAccount` are namespace-scoped.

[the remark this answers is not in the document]
Miciah: Printing the stringified `types.NamespacedName` value is simpler if you have a `types.NamespacedName` value, but yeah, this doesn't apply with a `*operatorv1alpha1.Contour` value.

controller/contour/controller.go
```diff
@@ -112,12 +112,18 @@ func (r *Reconciler) ensureContour(ctx context.Context, contour *operatorv1alpha
...
+	if err := r.ensureRBAC(ctx, contour); err != nil {
+		return fmt.Errorf("failed to ensure rbac for contour %s/%s: %w", contour.Namespace, contour.Name, err)
+	}
 	return nil
 }
 
 // ensureContourRemoved ensures all resources for the given contour do not exist
 // in namespace ns.
 func (r *Reconciler) ensureContourRemoved(ctx context.Context, contour *operatorv1alpha1.Contour, ns string) error {
+	if err := r.ensureRBACRemoved(ctx, contour); err != nil {
```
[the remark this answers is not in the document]
danehans: We do want to support multiple contours (in the same or different ns's). Thanks for catching this, I'll update.

danehans:
/hold

Still working to address the comments from @Miciah.

controller/contour/rbac.go
```diff
@@ -0,0 +1,263 @@
...
+	return nil
+}
+
+// ensureServiceAccount ensures a ServiceAccount resource exists in the provided ns/name.
+func (r *Reconciler) ensureServiceAccount(ctx context.Context, ns, name string) error {
+	sa := &corev1.ServiceAccount{ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: ns}}
+	err := r.Client.Get(ctx, types.NamespacedName{Namespace: ns, Name: name}, sa)
+	switch {
+	case err == nil:
+		r.Log.Info("service account exists; skipped adding", "namespace", sa.Namespace, "name", sa.Name)
```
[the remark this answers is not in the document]
danehans: @Miciah it's nice having the logging details during early dev, but I do agree that this level of logging should be removed in the near future. I created https://github.com/projectcontour/contour-operator/issues/28 to capture this issue.

controller/contour/rbac.go
```diff
@@ -0,0 +1,263 @@
...
+	secret := rbacv1.PolicyRule{
+		Verbs:     verbGLW,
+		APIGroups: groupAll,
+		Resources: []string{"secrets"},
+	}
+	svc := rbacv1.PolicyRule{
+		Verbs:     verbGLW,
+		APIGroups: groupAll,
+		Resources: []string{"services"},
+	}
```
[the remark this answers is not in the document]
danehans: @Miciah that does reduce the code, but the policy rules will slightly differ from what is generated by Contour in `examples/contour/02-rbac.yaml`. At this time, I prefer the rules are identical and start with condensing the upstream policy rules.

danehans:
@Miciah Thanks for the review. I believe I addressed all of your comments in commit `08e1012`, PTAL when you have a moment.

danehans:
@jpeach @skriss @stevesloka @Miciah PR is ready for review, PTAL.

controller/contour/rbac.go
```diff
@@ -0,0 +1,413 @@
...
+// the provided name.
+func desiredRole(name types.NamespacedName) *rbacv1.Role {
+	groupAll := []string{""}
+	verbCU := []string{"create", "update"}
+	secret := rbacv1.PolicyRule{
+		Verbs:     verbCU,
+		APIGroups: groupAll,
+		Resources: []string{"secrets"},
+	}
+	return &rbacv1.Role{
```
stevesloka, line 199: Should we overload the `Make` methods below to help create these as well?

controller/contour/rbac.go
```diff
@@ -0,0 +1,413 @@
...
+			Name:      name.Name,
+		},
+		Rules: []rbacv1.PolicyRule{secret},
+	}
+}
+
+// ensureRoleBinding ensures a RoleBinding resource with the provided name exists.
+// The RoleBinding will use svcAct for the subject and role for the role reference.
+func (r *Reconciler) ensureRoleBinding(ctx context.Context, name types.NamespacedName, svcAct *corev1.ServiceAccount, role *rbacv1.Role) error {
+	rb := &rbacv1.RoleBinding{
```
stevesloka, line 211: Use the `Make` methods?

controller/contour/rbac.go
```diff
@@ -0,0 +1,413 @@
...
+		return fmt.Errorf("failed to get role binding %s: %w", rb.Name, err)
+	}
+	r.Log.Info("role binding exists; skipped adding", "name", rb.Name)
+	return nil
+}
+
+// ensureClusterRoleBinding ensures a ClusterRoleBinding resource with the provided
+// name exists, using roleName for the role reference and svcAct for the subject.
+func (r *Reconciler) ensureClusterRoleBinding(ctx context.Context, name, roleName string, svcAct types.NamespacedName) error {
+	crb := &rbacv1.ClusterRoleBinding{
```
stevesloka, line 245: Same here with the make methods.

controller/contour/rbac.go
```diff
@@ -0,0 +1,413 @@
...
+
+// ensureRBACRemoved ensures all the necessary RBAC resources for the provided
+// contour do not exist.
+func (r *Reconciler) ensureRBACRemoved(ctx context.Context, contour *operatorv1alpha1.Contour) error {
+	var errs []error
+	ns := contour.Spec.Namespace.Name
+	objectsToDelete := []runtime.Object{}
+	exist, err := r.contoursExistWithSameSpecNs(ctx, contour)
+	if err != nil {
+		return fmt.Errorf("failed to verify if contours exist in namespace %s: %w",
```
stevesloka, line 283: nit: s/contours exist/contour exists/

controller/contour/rbac.go
```diff
@@ -0,0 +1,413 @@
...
+			contour.Spec.Namespace.Name, err)
+	}
+	if !exist {
```
stevesloka, line 286: Maybe I'm not reading right, but if the resources don't exist, we then attempt to delete? Is this just a way to make sure it's all gone if something is partially deleted?

controller/contour/controller.go
```diff
@@ -126,3 +132,33 @@ func (r *Reconciler) ensureContourRemoved(ctx context.Context, contour *operator
...
 }
+
+// contoursExist lists Contour objects in all namespaces, returning the list
+// and true if any exist other than contour.
+func (r *Reconciler) contoursExist(ctx context.Context, contour *operatorv1alpha1.Contour) (bool, *operatorv1alpha1.ContourList, error) {
+	contours := &operatorv1alpha1.ContourList{}
+	if err := r.Client.List(ctx, contours); err != nil {
+		return false, nil, fmt.Errorf("failed to list contours: %w", err)
+	}
+	if len(contours.Items) == 0 || len(contours.Items) == 1 && contours.Items[0].Name == contour.Name {
```
stevesloka, line 143: Maybe add a test for this? I had to read a couple times to see what was desired.

util/object.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// MakeClusterRole makes a ClusterRole object using the provided
+// name for the object's name.
+func MakeClusterRole(name string) *rbacv1.ClusterRole {
```
stevesloka, line 27: My first thought was that `Make` was actually creating the object in Kubernetes. I think it could work but would `Generate` be more clear? This is a super nit thing, so no biggie. Names are hard. 😢  Just thinking out loud.

controller/contour/controller.go
danehans: @stevesloka I'm working on https://github.com/projectcontour/contour-operator/pull/29 to add integration testing.

controller/contour/rbac.go
danehans:
```
exist, err := r.otherContoursExistInSpecNs(ctx, contour)
	if err != nil {
		return fmt.Errorf("failed to verify if contours exist in namespace %s: %w",
			contour.Spec.Namespace.Name, err)
	}
```

@stevesloka `otherContoursExistInSpecNs()` checks if any `contours` exist in the same ns, so I believe the error message is correct.

controller/contour/rbac.go
danehans: > if the resources don't exist, we then attempt to delete?

@stevesloka adding to https://github.com/projectcontour/contour-operator/pull/21#discussion_r494462378, `otherContoursExistInSpecNs()` returns a bool whether other `Contours` exist in the same ns as `contour`. If `false`, it's safe to remove ns-scoped RBAC resources. Similar logic is used later in the func for cluster-scoped RBAC resources.

util/object.go
danehans: I'll change the prefix Make -> Generate.

controller/contour/rbac.go
stevesloka: Ahh ok, it just reads funny to me to see "contours". =)

Might help to add some comments to this. I understand it now, you're first looking for any Contour instances in the namespace, then you look for any instances in any namespace. 

A future refactoring might make these rbac rules specific to the instance that got created. Possibly, one instance of Contour should only be able to have access to specific objects within specific namespaces.

controller/contour/controller.go
stevesloka: Integration testing is good, but unit tests are also super helpful to describe these types of logic.

danehans:
@stevesloka thanks for reviewing this PR. Your feedback is addressed in commit `0eafee5` or in the comments above, PTAL.

controller/contour/rbac.go
```diff
@@ -0,0 +1,372 @@
+package contour
```
stevesloka, line 1: This needs the standard Contour copyright header.

controller/contour/rbac.go
```diff
@@ -0,0 +1,372 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	operatorv1alpha1 "github.com/projectcontour/contour-operator/api/v1alpha1"
+	oputil "github.com/projectcontour/contour-operator/util"
+
```
stevesloka, line 9: nit: fixup the spacings here so the imports are together.

util/object.go
```diff
@@ -0,0 +1,91 @@
+/*
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
stevesloka, line 4: This should have the standard Contour header.

stevesloka (changes requested):
Just a couple cleanup nits @danehans and this is good to me. =)

util/object.go
danehans: @stevesloka this is a known [issue](https://github.com/projectcontour/contour-operator/issues/11). I'll fix the header across the entire project in a separate PR that closes the issue.

controller/contour/rbac.go
danehans: Same as https://github.com/projectcontour/contour-operator/pull/21#discussion_r498556284.

stevesloka (approved):
/lgtm

PR #22, opened 2020-09-14
Updates Replicas Field Requirement in Contour API

danehans:
Since the value of `Replicas` is being defaulted, the field should be required instead of optional.

/assign @jpeach @stevesloka 
/cc @Miciah


stevesloka (approved):
/lgtm

PR #23, opened 2020-09-14
Wraps Errors with %w Verb

danehans:
Wraps errors with `%w` instead of `%v`. In Go 1.13, `fmt.Errorf()` supports a new `%w` verb. When this verb is present, the error returned by `fmt.Errorf()` will have an Unwrap method returning the argument of `%w`, which must be an error. In all other ways, `%w` is identical to `%v`.

/assign @jpeach @stevesloka 
/cc @Miciah


stevesloka (approved):
/lgtm

PR #24, opened 2020-09-15
Updates json Tags for Defaulted API Fields

danehans:
Adds the `omitempty` json tag for defaulted API Fields. Previously, fields being defaulted did not contain `omitempty`. Without omitempty, an instance of `Contour` requires these fields which negates the desired defaulting behavior.

/assign @jpeach @stevesloka 
/cc @Miciah


api/v1alpha1/contour_types.go
```diff
@@ -49,13 +49,13 @@ type ContourSpec struct {
 	//
 	// +kubebuilder:default=2
 	// +kubebuilder:validation:Minimum=0
-	Replicas int32 `json:"replicas"`
+	Replicas int32 `json:"replicas,omitempty"`
```
stevesloka, line 52: I thought this meant to be required from a previous different PR?

danehans: @stevesloka the field is required in a sense. It's not required for a user to set, but it's required when the object is persisted. For example, a user can create a contour object with:
```
kind: Contour
metadata:
  name: contour-sample
spec: {}
```

which will be realized (thanks for kubebuilder defaulting) to:
```
kind: Contour
metadata:
  name: contour-sample
spec:
  replicas: 2
  namespace:
    name: projectcontour
    removeOnDeletion: false
status:
<SNIP>
```

api/v1alpha1/contour_types.go
```diff
@@ -49,13 +49,13 @@ type ContourSpec struct {
...
 
 	// namespace defines the schema of a Contour namespace.
 	// See each field for additional details.
 	//
-	// +optional
-	Namespace *NamespaceSpec `json:"namespace,omitempty"`
+	// +kubebuilder:default={name: "projectcontour", removeOnDeletion: false}
```
skriss, line 57: do we need to specify the defaults both here and on the fields within `NamespaceSpec`?

stevesloka (approved):
/lgtm

api/v1alpha1/contour_types.go
danehans: @skriss it's not required, but I think it's helpful when reading the code or the CRD. I can remove in a follow-on PR if you prefer.

skriss: not a big deal, i think it can stay as-is.

PR #19, opened 2020-09-02
Protects ensureNamespace() Against Race with Contour

danehans:
Previously, `ensureNamespace()` was returning an error during the 2nd pass of contour object reconciliation. This is due to a race between the Contour object being updated by ensureFinalizer() and the namespace create operation. This PR updates the control flow of the function to fix the error case for create and delete operations.

__Before:__
```
2020-09-02T12:33:22.826-0700  INFO  controllers.Contour reconciling {"request": "default/contour-sample"}
2020-09-02T12:33:22.838-0700  INFO  controllers.Contour added finalizer to contour  {"finalizer": "contour.operator.projectcontour.io/finalizer", "namespace": "default", "name": "contour-sample"}
2020-09-02T12:33:22.948-0700  INFO  controllers.Contour created namespace {"name": "projectcontour"}
2020-09-02T12:33:22.948-0700  DEBUG controller  Successfully Reconciled {"reconcilerGroup": "operator.projectcontour.io", "reconcilerKind": "Contour", "controller": "contour", "name": "contour-sample", "namespace": "default"}
2020-09-02T12:33:22.948-0700  INFO  controllers.Contour reconciling {"request": "default/contour-sample"}
2020-09-02T12:33:22.974-0700  ERROR controller  Reconciler error  {"reconcilerGroup": "operator.projectcontour.io", "reconcilerKind": "Contour", "controller": "contour", "name": "contour-sample", "namespace": "default", "error": "failed to ensure namespace projectcontour: failed to create namespace projectcontour: namespaces \"projectcontour\" already exists"}
github.com/go-logr/zapr.(*zapLogger).Error
  /Users/daneyonhansen/code/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
  /Users/daneyonhansen/code/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:237
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
  /Users/daneyonhansen/code/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.2/pkg/internal/controller/controller.go:209
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
[...]

PR #25, opened 2020-09-15
Implements Contour Namespace API

danehans:
Implements the Namespace API added in https://github.com/projectcontour/contour-operator/pull/20. Updates `ensureNamespace()` and `ensureNamespaceRemoved()` to use a `Contour` object instead of a namespace string as a parameter. Updates `ensureNamespaceRemoved()` to only remove the namespace if RemoveOnDeletion is unset and the namespace does not match a name in namespaceBlacklist.

Requires PR https://github.com/projectcontour/contour-operator/pull/24

/assign @jpeach @stevesloka 
/cc @Miciah

### DIFF
--- a/controller/contour/controller.go
+++ b/controller/contour/controller.go
@@ -71,7 +71,7 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// The contour is safe to process, so ensure current state matches desired state.
 	desired := contour.ObjectMeta.DeletionTimestamp.IsZero()
 	if desired {
-		if err := r.ensureContour(ctx, contour, defaultContourNamespace); err != nil {
+		if err := r.ensureContour(ctx, contour); err != nil {
 			switch e := err.(type) {
 			case retryable.Error:
 				r.Log.Error(e, "got retryable error; requeueing", "after", e.After())
@@ -81,7 +81,7 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			}
 		}
 	} else {
-		if err := r.ensureContourRemoved(ctx, contour, defaultContourNamespace); err != nil {
+		if err := r.ensureContourRemoved(ctx, contour); err != nil {
 			switch e := err.(type) {
 			case retryable.Error:
 				r.Log.Error(e, "got retryable error; requeueing", "after", e.After())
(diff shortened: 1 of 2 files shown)


controller/contour/namespace.go
```diff
@@ -20,20 +20,20 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
 )
 
-const (
-	// defaultContourNamespace is the name of the namespace
-	// used by Contour.
-	defaultContourNamespace = "projectcontour"
-)
+// namespaceBlacklist is a list of namespace names that should not be removed.
+var namespaceBlacklist = []string{"contour-operator", "default"}
```
stevesloka, line 32: Need a new name for this, maybe something like `coreNamespaceList` or something?

stevesloka: Also should add `kube-system` to this list or have other logic to not allow specific namespaces to be created?

stevesloka (changes requested):

controller/contour/namespace.go
danehans: @stevesloka thanks for the review. You're spot-on with both comments. I'll work on the updates and push a new commit.

danehans:
@stevesloka commit `aef797c` addresses your feedback and also implements a portion of the API that was previously missing (not removing the namespace if another contour exists), PTAL.

/cc @Miciah

controller/contour/controller.go
```diff
@@ -103,26 +103,47 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 // ensureContour ensures all necessary resources exist for the given contour
 // in namespace ns.
```
skriss, line 105: nit: the godoc comment still refers to `ns`.

controller/contour/controller.go
```diff
@@ -103,26 +103,47 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		return fmt.Errorf("failed to ensure namespace %s: %w", ns, err)
+	if err := r.ensureNamespace(ctx, contour); err != nil {
+		return fmt.Errorf("failed to ensure namespace %s for contour %s/%s: %w",
+			contour.Spec.Namespace.Name, contour.Namespace, contour.Name, err)
 	}
 	return nil
 }
 
 // ensureContourRemoved ensures all resources for the given contour do not exist
 // in namespace ns.
```
skriss, line 120: nit: the godoc comment still refers to `ns`.

controller/contour/namespace.go
```diff
@@ -50,21 +50,45 @@ func (r *Reconciler) ensureNamespace(ctx context.Context, name string) error {
...
-		if err := r.Client.Delete(ctx, ns); err != nil {
-			return fmt.Errorf("failed to delete namespace %s: %w", ns.Name, err)
-		}
-		r.Log.Info("deleted namespace", "name", ns.Name)
-		return nil
-	case errors.IsNotFound(err):
-		r.Log.Info("namespace does not exist; skipping removal", "name", ns.Name)
+// ensureNamespaceRemoved ensures the namespace for the provided contour is removed,
+// bypassing deletion if any of the following conditions apply:
+//   - RemoveOnDeletion is set.
```
skriss, line 55: should be "RemoveOnDeletion is **not** set", or "...set to false".

controller/contour/controller.go
```diff
@@ -103,26 +103,47 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	}
 	if err := r.ensureFinalizerRemoved(ctx, contour); err != nil {
 		return fmt.Errorf("failed to remove finalizer from contour %s/%s: %w", contour.Namespace, contour.Name, err)
 	}
 	return nil
 }
+
+// contoursExist lists Contour objects in all namespaces if all is true or only the
+// namespace of contour otherwise, returning true if any exist other than contour.
+func (r *Reconciler) contoursExist(ctx context.Context, contour *operatorv1alpha1.Contour, all bool) (bool, error) {
```
skriss, line 134: `all` is not currently specified as `true` anywhere, but I assume you have a plan to use it?

skriss, line 134: maybe `otherContoursExist` or something like that would be slightly more clear.

skriss (changes requested):
handful of comments but this mostly looks good!

controller/contour/controller.go
danehans: ack

danehans: I replaced this func with https://github.com/projectcontour/contour-operator/pull/21/files#diff-eee6f38bbe0411df081f7393625f732eR136-R159. I'll remove either based on which PR merges first.

controller/contour/namespace.go
danehans: ack

controller/contour/controller.go
danehans: ack

controller/contour/controller.go
danehans: ack

danehans:
@skriss I really appreciate the review. Commit `c3ac3d3` addresses your feedback, PTAL.

skriss:
Needs a rebase now that #19 has merged.

danehans:
@skriss commit `58ac6b6` resolves the conflict, PTAL.

skriss (approved):
LGTM :shipit:

danehans:
@stevesloka PTAL when you have a moment.

stevesloka (approved):

PR #27, opened 2020-09-17
Updates Docs for RemoveOnDeletion API Field

danehans:
Updates the `RemoveOnDeletion` API field docs to include "kube-system" as a namespace that will __not__ be removed even when a `true` value is provided.

See https://github.com/projectcontour/contour-operator/pull/25#discussion_r489467685 for additional background.

/assign @stevesloka 
/cc @Miciah


jpeach (approved):
LGTM

PR #30, opened 2020-09-25
Adds Contour ConfigMap Support

danehans:
Adds support for managing Contour's ConfigMap.

Fixes https://github.com/projectcontour/contour-operator/issues/38

@jpeach @skriss @stevesloka @Miciah PTAL

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -12,7 +12,9 @@ rules:
   - configmaps
   verbs:
   - create
+  - delete
   - get
+  - list
   - update
 - apiGroups:
   - ""
--- a/controller/contour/configmap.go
+++ b/controller/contour/configmap.go
@@ -1,0 +1,233 @@
+/*
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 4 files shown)


jpeach:
I don't think the template is the right approach here. Contour should expose a package with the Go struct for the config file, and we can consume that directly here.

jpeach:
xref https://github.com/projectcontour/contour/issues/2621

stevesloka:
The alternative would be to move the the CRD approach suggested here (https://github.com/projectcontour/contour/issues/2924#issuecomment-695081070), but is not built out just yet.

danehans:
@jpeach is it possible to proceed with this approach since this is what's implemented today? I can create an issue that xrefs https://github.com/projectcontour/contour/issues/2621 and https://github.com/projectcontour/contour/issues/2924#issuecomment-695081070 and update the contour-operator when these projectcontour/contour issues are resolved.

stevesloka:
@danehans I think this is fine to move forward with. The CRD swap is a lot more work and will require a deprecation window that shouldn't hold up this project. =)

danehans:
@stevesloka I've updated the PR based, PTAL.

/cc @jpeach @Miciah

jpeach (approved):

PR #31, opened 2020-09-28
Adds Certgen Job Support

danehans:
Adds support for managing the Contour certgen `Job`.

Fixes https://github.com/projectcontour/contour-operator/issues/39
 
@jpeach @skriss @stevesloka @Miciah PTAL

### DIFF
--- a/controller/contour/controller.go
+++ b/controller/contour/controller.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -62,6 +63,7 @@ type Reconciler struct {
 // +kubebuilder:rbac:groups=projectcontour.io,resources=httpproxies/status,verbs=create;get;update
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings;roles;rolebindings,verbs=get;list;delete;create;update;watch
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;delete;create
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;delete;create;update
 
 func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	ctx := context.Background()
@@ -133,11 +135,17 @@ func (r *Reconciler) ensureContour(ctx context.Context, contour *operatorv1alpha
 	if err := r.ensureConfigMap(ctx, contour); err != nil {
 		return fmt.Errorf("failed to ensure configmap for contour %s/%s: %w", contour.Namespace, contour.Name, err)
 	}
+	if err := r.ensureJob(ctx, contour); err != nil {
+		return fmt.Errorf("failed to ensure job for contour %s/%s: %w", contour.Namespace, contour.Name, err)
+	}
 	return nil
 }
 
 // ensureContourRemoved ensures all resources for the given contour do not exist.
 func (r *Reconciler) ensureContourRemoved(ctx context.Context, contour *operatorv1alpha1.Contour) error {
+	if err := r.ensureJobDeleted(ctx, contour); err != nil {
+		return fmt.Errorf("failed to remove job from contour %s/%s: %w", contour.Namespace, contour.Name, err)
+	}
 	if err := r.ensureConfigMapDeleted(ctx, contour); err != nil {
 		return fmt.Errorf("failed to remove configmap for contour %s/%s: %w",
 			contour.Namespace, contour.Name, err)
@@ -183,4 +191,11 @@ func (r *Reconciler) otherContoursExistInSpecNs(ctx context.Context, contour *op
 		}
 	}
 	return false, nil
+}
+
+// contourOwningSelector returns a label selector based on the provided contour.
+func contourOwningSelector(contour *operatorv1alpha1.Contour) *metav1.LabelSelector {
+	return &metav1.LabelSelector{
+		MatchLabels: map[string]string{owningContourLabel: contour.Name},
+	}
 }
--- a/controller/contour/job.go
+++ b/controller/contour/job.go
@@ -1,0 +1,221 @@
+/*
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package contour
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	operatorv1alpha1 "github.com/projectcontour/contour-operator/api/v1alpha1"
+	oputil "github.com/projectcontour/contour-operator/util"
+	utilequality "github.com/projectcontour/contour-operator/util/equality"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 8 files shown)


jpeach:
Some meta-comments on the approach:

1. It would be great if there was a Spec option to select whether certgen or cert-manager generates the secrets.
2. certgen should probably run with a unique (generated?) job name
3. IIUC most clusters don't have the necessary feature gate turned on, so the operator could delete the certgen job if it is successful
4. Does the operator even need to run certgen? It could just generate secrets directly?
5. Interested about your thoughts on keeping stuff like this in sync with the Contour deployment YAML?

stevesloka:
I think the operator should just create the certs without running a job, but that package isn't exposed in Contour so we need to do that work first. Duplicating the code doesn't seem like a good approach, but would unblock the operator work. I guess it could also call the contour binary like the certgen job as an another intermediate step.

jpeach:
> I think the operator should just create the certs without running a job,

Yep, I agree that's a good option. However, I'd also like the operator to allow the possibility of using cert-manager to create the certificates, which implies some config in the contour spec WDYT @danehans ?

danehans:
@jpeach My plan for the operator is to be consistent with `https://projectcontour.io/quickstart/contour.yaml` and then iterate from there. This will help in setting a baseline and attracting additional contributors. I do like the idea of providing cert gen options and a spec field will be need to expose the options. I think this can be a nice feature to distinguish the operator from other Contour management approaches. Are you OK with this direction? If so, I'll create an issue for managing cert generation.

jpeach:
> @jpeach My plan for the operator is to be consistent with `https://projectcontour.io/quickstart/contour.yaml` and then iterate from there. This will help in setting a baseline and attracting additional contributors. 

I see, that sounds reasonable to me. Let's capture this approach in the code comments?

stevesloka:
The operator should be able to generate certs for a user, this is super helpful to get running quickly as well as fill that gap for those who don't want to deal with their own cert generation, but it is important to allow certs to come from cert-manager, but also to come from standard k8s secrets directly should a user have their own internal system for generating certs.

controller/contour/job.go
```diff
@@ -0,0 +1,197 @@
...
+	changed, updated := jobConfigChanged(current, desired)
+	if !changed {
+		return nil
+	}
+
+	// Diff before updating because the client may mutate the object.
+	diff := cmp.Diff(current, updated, cmpopts.EquateEmpty())
+	if err := r.Client.Update(ctx, updated); err != nil {
+		return fmt.Errorf("failed to update job %s/%s: %w", updated.Namespace, updated.Name, err)
+	}
```
jpeach, line 183: Most job fields are read-only. We probably need to just delete the job here.

controller/contour/job.go
```diff
@@ -0,0 +1,197 @@
...
+	r.Log.Info("updated job", "namespace", updated.Namespace, "name", updated.Name, "diff", diff)
+	return nil
+}
+
+// jobConfigChanged checks if current and expected match and if not,
+// returns false and the expect job.
+func jobConfigChanged(current, expected *batchv1.Job) (bool, *batchv1.Job) {
+	if cmp.Equal(current.Spec, expected.Spec, cmpopts.EquateEmpty()) {
```
jpeach, line 191: Prefer semantic equality?

https://godoc.org/k8s.io/apimachinery/pkg/api/equality

controller/contour/job.go
```diff
@@ -0,0 +1,197 @@
...
+		if err := r.updateJob(ctx, current, desired); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// currentJob gets the Job for contour from the api server, returning true
+// and the Job if successful.
+func (r *Reconciler) currentJob(ctx context.Context, contour *operatorv1alpha1.Contour) (bool, *batchv1.Job, error) {
```
jpeach, line 75: Rather than trying to update the job in place, what I'd suggest is using GenerateName to name the job (IIRC that can be made to work) and labelling it with an owner ref. That way you can always List() to find the certgen jobs you own, and still generate unique names when you need a new one.

controller/contour/job.go
```diff
@@ -0,0 +1,197 @@
...
+
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      contour.Name + "-certgen",
+			Namespace: contour.Spec.Namespace.Name,
+		},
+		Spec: batchv1.JobSpec{
+			Parallelism:  &one,
+			Completions:  &one,
+			BackoffLimit: &one,
```
jpeach, line 104: You can use https://godoc.org/k8s.io/utils/pointer#Int32Ptr here.

controller/contour/job.go
```diff
@@ -0,0 +1,197 @@
...
+			},
+		},
+	}
+
+	// TODO [danehans]: Once https://github.com/projectcontour/contour-operator/pull/30/files
+	// merges, associate contour controller owning label.
+	/*job.Labels = map[string]string{
+		// associate the job with the contour
+		controllerJobLabel: contour.Name,
+	}*/
```
jpeach, line 123: :)

controller/contour/controller.go
```diff
@@ -112,6 +113,9 @@ func (r *Reconciler) ensureContour(ctx context.Context, contour *operatorv1alpha
 		return fmt.Errorf("failed to ensure namespace %s for contour %s/%s: %w",
 			contour.Spec.Namespace.Name, contour.Namespace, contour.Name, err)
 	}
+	if err := r.ensureJob(ctx, contour); err != nil {
```
jpeach, line 116: The real dependency here is whether the TLS secrets are present, so perhaps we should check for those first, then use certgen as a secret generating strategy. I don't mind leaving this as is for now, but add some comments.

jpeach (commented):
Added some initial feedback.

controller/contour/job.go
danehans: @jpeach the [k8s docs](https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup) state:
> When a Job completes, no more Pods are created, but the Pods are not deleted either. Keeping them around allows you to still view the logs of completed pods to check for errors, warnings, or other diagnostic output.

It seems like keeping the job around does provide value. Should the operator simply ignore any changes to the job?

controller/contour/job.go
danehans: > labelling it with an owner ref.

@jpeach the `Contour` that creates the `Job` resides in a different namespace, so it can't be used as an owner. Is the original job the owner?

controller/contour/controller.go
danehans: added code docs.

danehans:
@jpeach I've updated the PR based on your feedback, PTAL.

controller/contour/job.go
jpeach: We hit this problem with the example deployment. We used to just ship YAML with the job spec and when you reapplied it (with the updated image field), changing the job spec would fail because job fields are read-only.

You can keep the job around if you like, but you can't update it. I see that you changed to `recreateJob`, which looks like the right approach.

controller/contour/job.go
```diff
@@ -0,0 +1,274 @@
...
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
+)
+
+const (
+	jobContainerName = "contour"
+	jobImageName     = "docker.io/projectcontour/contour:main"
```
jpeach, line 37: The job should use the versioned contour container image. The reason we used main historically was that re-applying the YAML would fail on upgrades because that would attempt to modify the job spec. I don't think you gave that issue here since you are recreating the job rather than modifying it.

danehans: @jpeach I updated the desired `Job` to use the image from the `Config` API and removed `jobImageName`. I would expect the image from `Config` is set as part of the release process. If not, then the default image tag should be updated when the operator is ready for release.

controller/contour/job.go
```diff
@@ -0,0 +1,221 @@
...
+		Image:           image,
+		ImagePullPolicy: corev1.PullAlways,
+		Command: []string{
+			"contour",
+			"certgen",
+			"--kube",
+			"--incluster",
+			"--overwrite",
+			"--secrets-format=compact",
+			fmt.Sprintf("--namespace=$(%s)", jobNsEnvVar),
```
jpeach, line 133:
```suggestion
			fmt.Sprintf("--namespace=%s", jobNsEnvVar),
```

jpeach (approved):

PR #33, opened 2020-09-30
Adds Contour Deployment Support

danehans:
Adds support for managing the Contour `Deployment` based on [this](https://github.com/projectcontour/contour/blob/main/examples/contour/03-contour.yaml) reference.

Fixes https://github.com/projectcontour/contour-operator/issues/37

/cc @jpeach @skriss @stevesloka @Miciah

### DIFF
--- a/controller/contour/deployment.go
+++ b/controller/contour/deployment.go
@@ -1,0 +1,351 @@
+/*
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package contour
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	operatorv1alpha1 "github.com/projectcontour/contour-operator/api/v1alpha1"
+	utilequality "github.com/projectcontour/contour-operator/util/equality"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
+)
+
+const (
+	// contourContainerName is the name of the Contour container.
+	contourContainerName = "contour"
+	// contourCfgMapName is the name of the Contour configmap.
+	contourCfgMapName = "contour"
+	// contourNsEnvVar is the name of the contour namespace environment variable.
+	contourNsEnvVar = "CONTOUR_NAMESPACE"
+	// contourPodEnvVar is the name of the contour pod name environment variable.
+	contourPodEnvVar = "POD_NAME"
+	// contourCertsVolName is the name of the contour certificates volume.
+	contourCertsVolName = "contourcert"
+	// contourCertsVolMntDir is the directory name of the contour certificates volume.
+	contourCertsVolMntDir = "certs"
+	// contourCertsSecretName is the name of the secret used as the certificate volume source.
+	contourCertsSecretName = contourCertsVolName
+	// contourCfgVolName is the name of the contour configuration volume.
+	contourCfgVolName = "contour-config"
+	// contourCfgVolMntDir is the directory name of the contour configuration volume.
+	contourCfgVolMntDir = "config"
+	// contourCfgFileName is the name of the contour configuration file.
+	contourCfgFileName = "contour.yaml"
+	// contourDeploymentLabel identifies a deployment as a contour deployment,
+	// and the value is the name of the owning contour.
+	contourDeploymentLabel = "contour.operator.projectcontour.io/deployment-contour"
+)
+
+// ensureDeployment ensures a deployment exists for the given contour.
+func (r *Reconciler) ensureDeployment(ctx context.Context, contour *operatorv1alpha1.Contour) error {
+	desired, err := DesiredDeployment(contour, r.Config.Image)
+	if err != nil {
+		return fmt.Errorf("failed to build deployment: %w", err)
+	}
+
+	current, err := r.currentDeployment(ctx, contour)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			if err := r.createDeployment(ctx, desired); err != nil {
+				return fmt.Errorf("failed to create deployment %s/%s: %w", desired.Namespace, desired.Name, err)
+			}
+			return nil
+		}
+		return fmt.Errorf("failed to get deployment %s/%s: %w", desired.Namespace, desired.Name, err)
+	}
+
+	if err := r.updateDeploymentIfNeeded(ctx, current, desired); err != nil {
+		return fmt.Errorf("failed to update deployment %s/%s: %w", desired.Namespace, desired.Name, err)
+	}
+
+	return nil
+}
+
+// ensureDeploymentDeleted ensures the deployment for the provided contour
+// is deleted.
+func (r *Reconciler) ensureDeploymentDeleted(ctx context.Context, contour *operatorv1alpha1.Contour) error {
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: contour.Spec.Namespace.Name,
+			Name:      contour.Name,
+		},
+	}
+
+	if err := r.Client.Delete(ctx, deployment); err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+	r.Log.Info("deleted deployment", "namespace", deployment.Namespace, "name", deployment.Name)
+
+	return nil
+}
+
+// DesiredDeployment returns the desired deployment for the provided contour using
+// image as Contour's container image.
+func DesiredDeployment(contour *operatorv1alpha1.Contour, image string) (*appsv1.Deployment, error) {
+	parsedImage := strings.Split(image, ":")
+	labels := map[string]string{
+		"app.kubernetes.io/name":     "contour",
+		"app.kubernetes.io/instance": contour.Name,
+		// The image tag is used as the version.
+		"app.kubernetes.io/version":    parsedImage[1],
+		"app.kubernetes.io/component":  "ingress-controller",
+		"app.kubernetes.io/managed-by": "contour-operator",
+		// Associate the deployment with the provided contour.
+		operatorv1alpha1.OwningContourLabel: contour.Name,
+	}
+
+	container := corev1.Container{
+		Name:            contourContainerName,
+		Image:           image,
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		Command: []string{
+			"contour",
+		},
+		Args: []string{
+			"serve",
+			"--incluster",
+			"--xds-address=0.0.0.0",
+			"--xds-port=8001",
+			"--envoy-service-http-port=80",
+			"--envoy-service-https-port=443",
+			fmt.Sprintf("--contour-cafile=%s", filepath.Join("/", contourCertsVolMntDir, "ca.crt")),
+			fmt.Sprintf("--contour-cert-file=%s", filepath.Join("/", contourCertsVolMntDir, "tls.crt")),
+			fmt.Sprintf("--contour-key-file=%s", filepath.Join("/", contourCertsVolMntDir, "tls.key")),
+			fmt.Sprintf("--config-path=%s", filepath.Join("/", contourCfgVolMntDir, contourCfgFileName)),
+		},
+		Env: []corev1.EnvVar{
+			{
+				Name: contourNsEnvVar,
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{
+						APIVersion: "v1",
+						FieldPath:  "metadata.namespace",
+					},
+				},
+			},
+			{
+				Name: contourPodEnvVar,
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{
+						APIVersion: "v1",
+						FieldPath:  "metadata.name",
+					},
+				},
+			},
+		},
+		Ports: []corev1.ContainerPort{
+			{
+				Name:          "xds",
+				ContainerPort: 8001,
+				Protocol:      "TCP",
+			},
+			{
+				Name:          "debug",
+				ContainerPort: 8000,
+				Protocol:      "TCP",
+			},
+		},
+		LivenessProbe: &corev1.Probe{
+			Handler: corev1.Handler{
+				HTTPGet: &corev1.HTTPGetAction{
+					Scheme: corev1.URISchemeHTTP,
+					Path:   "/healthz",
+					Port:   intstr.IntOrString{IntVal: int32(8000)},
+				},
+			},
+			TimeoutSeconds:   int32(1),
+			PeriodSeconds:    int32(10),
+			SuccessThreshold: int32(1),
+			FailureThreshold: int32(3),
+		},
+		ReadinessProbe: &corev1.Probe{
+			Handler: corev1.Handler{
+				TCPSocket: &corev1.TCPSocketAction{
+					Port: intstr.IntOrString{
+						IntVal: int32(8001),
+					},
+				},
+			},
+			TimeoutSeconds:      int32(1),
+			InitialDelaySeconds: int32(15),
+			PeriodSeconds:       int32(10),
+			SuccessThreshold:    int32(1),
+			FailureThreshold:    int32(3),
+		},
+		TerminationMessagePath:   "/dev/termination-log",
+		TerminationMessagePolicy: corev1.TerminationMessageReadFile,
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      contourCertsVolName,
+				MountPath: filepath.Join("/", contourCertsVolMntDir),
+				ReadOnly:  true,
+			},
+			{
+				Name:      contourCfgVolName,
+				MountPath: filepath.Join("/", contourCfgVolMntDir),
+				ReadOnly:  true,
+			},
+		},
+	}
+
+	pointerTo := func(ios intstr.IntOrString) *intstr.IntOrString { return &ios }
+	deploy := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: contour.Spec.Namespace.Name,
+			Name:      contour.Name,
+			Labels:    labels,
+		},
+		Spec: appsv1.DeploymentSpec{
+			ProgressDeadlineSeconds: pointer.Int32Ptr(int32(600)),
+			Replicas:                &contour.Spec.Replicas,
+			RevisionHistoryLimit:    pointer.Int32Ptr(int32(10)),
+			// Ensure the deployment adopts only its own pods.
+			Selector: contourDeploymentPodSelector(contour),
+			Strategy: appsv1.DeploymentStrategy{
+				Type: appsv1.RollingUpdateDeploymentStrategyType,
+				RollingUpdate: &appsv1.RollingUpdateDeployment{
+					MaxSurge:       pointerTo(intstr.FromString("50%")),
+					MaxUnavailable: pointerTo(intstr.FromString("25%")),
+				},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					// TODO [danehans]: Remove the prometheus annotations when Contour is updated to
+					// show how the Prometheus Operator is used to scrape Contour/Envoy metrics.
+					Annotations: map[string]string{
+						"prometheus.io/scrape": "true",
+						"prometheus.io/port":   "8000",
+					},
+					Labels: contourDeploymentPodSelector(contour).MatchLabels,
+				},
+				Spec: corev1.PodSpec{
+					// TODO [danehans]: Readdress anti-affinity when https://github.com/projectcontour/contour/issues/2997
+					// is resolved.
+					Affinity: &corev1.Affinity{
+						PodAffinity: &corev1.PodAffinity{
+							PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
+								{
+									Weight: int32(100),
+									PodAffinityTerm: corev1.PodAffinityTerm{
+										TopologyKey: "kubernetes.io/hostname",
+										LabelSelector: &metav1.LabelSelector{
+											MatchLabels: labels,
+										},
+									},
+								},
+							},
+						},
+					},
+					Containers: []corev1.Container{container},
+					Volumes: []corev1.Volume{
+						{
+							Name: contourCertsVolName,
+							VolumeSource: corev1.VolumeSource{
+								Secret: &corev1.SecretVolumeSource{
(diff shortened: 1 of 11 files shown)


controller/contour/deployment_test.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+func checkDeploymentHasEnvVar(t *testing.T, deploy *appsv1.Deployment, name string) {
+	t.Helper()
+
+	for _, envVar := range deploy.Spec.Template.Spec.Containers[0].Env {
+		if envVar.Name == name {
+			break
+		}
+	}
```
Miciah, line 29: The function doesn't return a value or raise an error.
```suggestion
			return
		}
	}
	t.Errorf("deployment is missing environment variable %q", name)
```

controller/contour/deployment_test.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func checkContainerHasImage(t *testing.T, container *corev1.Container, image string) {
+	t.Helper()
+
+	if container.Image == image {
+		return
+	} else {
+		t.Errorf("deployment has invalid image %q", container.Image)
+	}
```
Miciah, line 56:
```suggestion
		return
	}
	t.Errorf("deployment has invalid image %q", container.Image)
```

controller/contour/deployment_test.go
```diff
@@ -0,0 +1,142 @@
...
+			Name:      testName,
+			Namespace: testNs,
+		},
+	}
+	deploy, err := desiredDeployment(ctr)
+	if err != nil {
+		t.Errorf("invalid deployment: %v", err)
+	}
+
+	container := checkDeploymentHasContainer(t, deploy, deployContainerName, true)
```
Miciah, line 71: Need a nil check, or need to change `checkDeploymentHasContainer` to use `t.Fatalf` instead of `t.Errorf`.

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const (
+	// TODO [danehans]: Once https://github.com/projectcontour/contour-operator/pull/30
+	// merges, associate contour controller owning label. The deployment will not get removed without this.
+	// ControllerDeploymentLabel identifies a deployment as a contour deployment,
```
Miciah, line 24:
```suggestion
	// controllerDeploymentLabel identifies a deployment as a contour deployment,
```

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+	// and the value is the name of the owning contour.
+	//controllerDeploymentLabel = "contour.operator.projectcontour.io/deployment-contour"
```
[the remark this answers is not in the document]
danehans: > Can you have two contours with the same name in distinct namespaces?

@Miciah I don't see why not. Let's say I want a "default" contour for test/dev/prod namespaces. Each namespace will have a `Contour` named "default" and a `Deployment` labeled "contour.operator.projectcontour.io/deployment-contour". What do you see as the benefit of the name being a ns/name for namespace-scoped resources?

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+		MountPath: certsVolumeMountPath,
+		ReadOnly:  true,
+	}
+	cfgMapPerms := int32(0644)
+	cfgVol := corev1.Volume{
+		Name: cfgVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: "contour",
```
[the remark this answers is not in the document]
danehans: ack

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// ensureContourDeleted ensures the Deployment resource for the provided
+// contour is deleted.
+func (r *Reconciler) ensureContourDeleted(ctx context.Context, contour *operatorv1alpha1.Contour) error {
```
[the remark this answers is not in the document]
danehans: ack

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+		}
+	}
+
+	r.Log.Info("deleted deployment", "namespace", deployment.Namespace, "name", deployment.Name)
+	return nil
+}
+
+// desiredDeployment returns the desired deployment for the provided contour.
+func desiredDeployment(contour *operatorv1alpha1.Contour) (*appsv1.Deployment, error) {
+	label := map[string]string{"app": contour.Name}
```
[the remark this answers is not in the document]
danehans: @Miciah I was trying to be consistent with projectcontour/contour: https://github.com/projectcontour/contour/blob/main/examples/contour/03-contour.yaml. 

https://github.com/projectcontour/contour/issues/1821 appears to track the issue that you mention. I'll update the labels accordingly.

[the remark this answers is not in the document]
jpeach: Happy to drop the "app" label in favor of well-known standards :)

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+			Selector: deploymentPodSelector(contour),
+		},
+	}
+
+	// TODO: Uncomment when https://github.com/projectcontour/contour-operator/pull/30 merges.
+	//deploy.SetOwnerReferences([]metav1.OwnerReference{newContourOwnerRef(contour)})
+
+	deploy.Spec.Template.Labels = deploymentPodSelector(contour).MatchLabels
+
+	annotations := map[string]string{"prometheus.io/scrape": "true", "prometheus.io/port": "8000"}
```
stevesloka, line 125: A different TODO, but we need to update Contour to show how to use the Prometheus Operator to scrape Contour/Envoy metrics. Doing that then wouldn't require this annotation.

stevesloka (commented):
@danehans I assume this is just your first pass to get it up and running, but many of the values should be able to be customized via the CRD, like image, pull policy, args to contour, etc.

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &contour.Spec.Replicas,
+			// Ensure the deploy adopts only its own pods.
+			Selector: deploymentPodSelector(contour),
+		},
+	}
+
+	// TODO: Uncomment when https://github.com/projectcontour/contour-operator/pull/30 merges.
+	//deploy.SetOwnerReferences([]metav1.OwnerReference{newContourOwnerRef(contour)})
```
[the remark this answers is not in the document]
danehans: @Miciah ah.. yes the ownerRef must be in the same ns. I'll use an owningRef for the deployment to own operand resources such as the configmap and use labels to associate the `Contour`<>`Deployment`.

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+		RollingUpdate: &appsv1.RollingUpdateDeployment{
+			MaxSurge: pointerTo(intstr.FromString("50%")),
+			// MaxUnavailable defaults to 25%
+		},
+	}
+
+	ctrContainer := corev1.Container{
+		Name:            deployContainerName,
+		Image:           deployImageName,
+		ImagePullPolicy: corev1.PullAlways,
```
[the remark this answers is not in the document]
danehans: I agree. However, I'm trying to match [upstream](https://github.com/projectcontour/contour/blob/main/examples/contour) as closely as possible for the 1st iteration of the operator. I created https://github.com/projectcontour/contour/issues/2994 to address this issue in projectcontour/contour.

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+			"serve",
+			"--incluster",
+			"--xds-address=0.0.0.0",
+			"--xds-port=8001",
+			"--envoy-service-http-port=80",
+			"--envoy-service-https-port=443",
+			fmt.Sprintf("--contour-cafile=%s/ca.crt", certsVolumeMountPath),
+			fmt.Sprintf("--contour-cert-file=%s/tls.crt", certsVolumeMountPath),
+			fmt.Sprintf("--contour-key-file=%s/tls.key", certsVolumeMountPath),
+			fmt.Sprintf("--config-path=%s/contour.yaml", cfgVolumeMountPath),
```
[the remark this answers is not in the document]
danehans: ack

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+	nonRoot := true
+	secCtx := corev1.PodSecurityContext{
+		RunAsUser:    &user,
+		RunAsGroup:   &group,
+		RunAsNonRoot: &nonRoot,
+	}
+	deploy.Spec.Template.Spec.SecurityContext = &secCtx
+
+	deploy.Spec.Template.Spec.Affinity = &corev1.Affinity{
+		PodAffinity: &corev1.PodAffinity{
```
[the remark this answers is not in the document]
danehans: Same as https://github.com/projectcontour/contour-operator/pull/33#discussion_r500412999.

danehans:
> @danehans I assume this is just your first pass to get it up and running, but many of the values should be able to be customized via the CRD, like image, pull policy, args to contour, etc.

@stevesloka that is correct. The first iteration of the operator will implement the [reference example](https://github.com/projectcontour/contour/blob/main/examples/contour).

controller/contour/deployment.go
danehans: @stevesloka do you have a contour/issue to xref?

controller/contour/deployment.go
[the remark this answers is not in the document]
danehans: @Miciah xref https://github.com/projectcontour/contour/issues/2997 for anti-affinity.

danehans:
@Miciah I've updated the PR based on your comments, PTAL when you have a moment.

controller/contour/deployment.go
```diff
@@ -0,0 +1,423 @@
...
+		return false, fmt.Errorf("failed to update deployment %s/%s: %v", updated.Namespace, updated.Name, err)
+	}
+	r.Log.Info("updated deployment", "namespace", updated.Namespace, "name", updated.Name)
+
+	return true, nil
+}
+
+// deploymentConfigChanged checks if current matches expected,
+// and if not returns the updated Deployment resource.
+func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
```
jpeach, line 346: I'm not sure that I understand the need for all the checks here. Can we just compare the spec and the labels?

```Go
updated := current.DeepCopy()

if apiequality.Semantic.DeepEqual(current.Spec, expected.Spec) {
    changed = true
    expected.Spec.DeepCopyInto(&updated.Spec)
}

for _, l := range expected.Labels {
   // If label doesn't exist in `updated`, add it
}
```

jpeach, line 346: The bool out parameter should come last, to make the caller more idiomatic:

```Go
deployment, ok := deploymentConfigChanged(current, expected); if ok {
    // do the update.
}
```

controller/contour/deployment.go
```diff
@@ -0,0 +1,423 @@
...
+// using version as the "app.kubernetes.io/version" label value.
+func desiredDeployment(contour *operatorv1alpha1.Contour, version string) (*appsv1.Deployment, error) {
+	labels := map[string]string{
+		"app.kubernetes.io/name":       "contour",
+		"app.kubernetes.io/instance":   contour.Name,
+		"app.kubernetes.io/version":    version,
+		"app.kubernetes.io/component":  "ingress-controller",
+		"app.kubernetes.io/part-of":    "project-contour",
+		"app.kubernetes.io/managed-by": "contour-operator",
+		// associate the deployment with the provided contour.
```
jpeach, line 127:
```suggestion
		// Associate the deployment with the provided contour.
```

controller/contour/deployment.go
```diff
@@ -0,0 +1,423 @@
...
+
+// desiredDeployment returns the desired deployment for the provided contour
+// using version as the "app.kubernetes.io/version" label value.
+func desiredDeployment(contour *operatorv1alpha1.Contour, version string) (*appsv1.Deployment, error) {
+	labels := map[string]string{
+		"app.kubernetes.io/name":       "contour",
+		"app.kubernetes.io/instance":   contour.Name,
+		"app.kubernetes.io/version":    version,
+		"app.kubernetes.io/component":  "ingress-controller",
+		"app.kubernetes.io/part-of":    "project-contour",
```
jpeach, line 125: `part-of` has some interpretation. I'd suggest `projectcontour.io` to refer to Contour the project, or maybe the name of the `Contour` object that is owning this controller?

controller/contour/deployment.go
```diff
@@ -0,0 +1,423 @@
...
+			Namespace: key.Namespace,
+			Name:      key.Name,
+		},
+	}
+
+	if err := r.Client.Delete(ctx, deployment); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+		return nil
```
jpeach, line 110: Why return nil rather than the deletion error?

controller/contour/deployment.go
```diff
@@ -0,0 +1,423 @@
...
+	if err := r.Client.Create(ctx, deploy); err != nil {
+		return fmt.Errorf("failed to create deployment %s/%s: %v", deploy.Namespace, deploy.Name, err)
+	}
+	r.Log.Info("created deployment", "namespace", deploy.Namespace, "name", deploy.Name)
+
+	return nil
+}
+
+// updateDeployment updates a Deployment if current does not match desired.
+func (r *Reconciler) updateDeployment(ctx context.Context, current, desired *appsv1.Deployment) (bool, error) {
```
jpeach, line 330: Why does this need to return the bool as well? Seems like we don't need to care whether there was an update performed, as long as we know the desired state was applied.

jpeach (changes requested):

util/strings/strings.go
```diff
@@ -0,0 +1,39 @@
...
+package strings
+
+import (
+	"fmt"
+	"strings"
+)
+
+// VersionFromImage parses image, returning the version or an
+// error if image can;t be parsed or doesn't contain a version.
+func VersionFromImage(image string) (string, error) {
```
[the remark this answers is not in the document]
danehans: I created https://github.com/projectcontour/contour-operator/issues/41 to address the need for image validation. In the meantime, I refactored the func to get the version (i.e. image tag) by splitting the "docker.io/projectcontour/contour:main" string.

controller/contour/deployment.go
```diff
@@ -0,0 +1,423 @@
...
+)
+
+// ensureDeployment ensures a deployment exists for the given contour.
+func (r *Reconciler) ensureDeployment(ctx context.Context, contour *operatorv1alpha1.Contour) error {
+	haveDepl, current, err := r.currentDeployment(ctx, contour)
+	if err != nil {
+		return err
+	}
+
+	version, err := stringsutil.VersionFromImage(deployImageName)
```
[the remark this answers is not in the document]
danehans: I agree and will update in https://github.com/projectcontour/contour-operator/issues/41 to address image validation.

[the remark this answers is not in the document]
danehans: Ack, I also updated the contour-operator cmd `--image` flag default tag from `latest` > `main`.

controller/contour/deployment.go
```diff
@@ -0,0 +1,423 @@
...
+		Type: appsv1.RollingUpdateDeploymentStrategyType,
+		RollingUpdate: &appsv1.RollingUpdateDeployment{
+			MaxSurge: pointerTo(intstr.FromString("50%")),
+			// MaxUnavailable defaults to 25%
+		},
+	}
+
+	ctrContainer := corev1.Container{
+		Name:            deployContainerName,
+		Image:           deployImageName,
```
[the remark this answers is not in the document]
danehans: ack

controller/contour/deployment.go
danehans: We already have "app.kubernetes.io/instance" and ""contour.operator.projectcontour.io/owning-contour" that refer to the Contour name. I'll remove this label for now until the community settles on a spec.

controller/contour/deployment.go
danehans: `ensureDeploymentDeleted()` is ensuring the deployment for the provided contour is deleted. If the error returned by the client delete operation is `NotFound` then the deployment is already deleted.

controller/contour/deployment.go
danehans: I originally went down this path because some spec fields were being defaulted by the api server instead of being specified by `desiredDeployment`. I'll update to compare spec and labels.

controller/contour/deployment.go
```diff
@@ -0,0 +1,423 @@
...
+	cfgVolumeName = "contour-config"
+	// cfgVolumeMountDir is the directory name of the contour configuration volume.
+	cfgVolumeMountDir = "config"
+	// cfgFileName is the name of the contour configuration file.
+	cfgFileName = "contour.yaml"
+)
+
+// ensureDeployment ensures a deployment exists for the given contour.
+func (r *Reconciler) ensureDeployment(ctx context.Context, contour *operatorv1alpha1.Contour) error {
+	haveDepl, current, err := r.currentDeployment(ctx, contour)
```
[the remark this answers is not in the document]
danehans: The problem with the above example is that `currentDeployment()` returns a bool so we know if a current `Deployment` exists or not and to either update the existing `Deployment` or create a new one.

danehans:
@jpeach I addressed your review feedback above and in commit `31d2afe`, PTAL when you have a moment.

controller/contour/deployment.go
```diff
@@ -0,0 +1,345 @@
...
+		},
+		Ports: []corev1.ContainerPort{
+			{
+				Name:          "xds",
+				ContainerPort: 8001,
+				Protocol:      "TCP",
+			},
+			{
+				Name:          "debug",
+				ContainerPort: 8000,
```
stevesloka, line 168: There really isn't a need to expose this port.

stevesloka (approved):
/lgtm

controller/contour/deployment.go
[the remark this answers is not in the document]
Miciah: My point was that if the value of the `contour.operator.projectcontour.io/deployment-contour` label is just the name of the contour, then to take your example, you would have 3 contours with `contour.operator.projectcontour.io/deployment-contour=default`.  I am suggesting that you change the label value to the _namespaced_ name of the contour, so you get "contour.operator.projectcontour.io/deployment-contour=test/default`, `contour.operator.projectcontour.io/deployment-contour=dev/default`, and `contour.operator.projectcontour.io/deployment-contour=prod/default`.

controller/contour/deployment.go
jpeach: Oh I misread it; I was expecting
```Go
		if errors.IsNotFound(err) {
			return nil
		}

		return err
```

controller/contour/deployment.go
```diff
@@ -0,0 +1,345 @@
...
+			Handler: corev1.Handler{
+				TCPSocket: &corev1.TCPSocketAction{Port: intstr.IntOrString{IntVal: int32(8001)}},
+			},
+			InitialDelaySeconds: int32(15),
+			PeriodSeconds:       int32(10),
+		},
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      certsVolumeName,
+				MountPath: certsVolumeMountDir,
```
Miciah, line 190:
```suggestion
				MountPath: filepath.Join("/", certsVolumeMountDir),
```
Or include the leading slash in `certsVolumeMountDir` itself.

controller/contour/deployment.go
```diff
@@ -0,0 +1,345 @@
...
+				ReadOnly:  true,
+			},
+			{
+				Name:      cfgVolumeName,
+				MountPath: cfgVolumeMountDir,
```
Miciah, line 195:
```suggestion
				MountPath: filepath.Join("/", cfgVolumeMountDir),
```

controller/contour/deployment.go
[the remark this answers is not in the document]
Miciah: Hm, but https://github.com/projectcontour/contour/blob/93aed23549fc87ad59689079da5cf0cd5bfaa3d2/examples/contour/03-contour.yaml#L27-L36 _does_ specify `podAntiAffinity`.  So (a) I think this PR should specify `podAntiAffinity` as well, and (b) it isn't clear to me what #2997 is suggesting be changed.

controller/contour/deployment.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+	r.Log.Info("updated deployment", "namespace", updated.Namespace, "name", updated.Name, "diff", diff)
+
+	return true, nil
+}
+
+// deploymentConfigChanged checks if current matches expected,
+// and if not returns the updated Deployment resource.
+func deploymentConfigChanged(current, expected *appsv1.Deployment) (bool, *appsv1.Deployment) {
+	if cmp.Equal(current.Spec, expected.Spec, cmpopts.EquateEmpty()) {
```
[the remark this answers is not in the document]
Miciah: I see you changed `cmp.Equal` to `apiequality.Semantic.DeepEqual`; does the latter ignore API-defaulted fields such as the ones I mentioned?

controller/contour/deployment.go
danehans: @stevesloka the port is being exposed in https://raw.githubusercontent.com/projectcontour/contour/release-1.9/examples/render/contour.yaml, so I'd like to keep it for consistency. I created https://github.com/projectcontour/contour-operator/issues/46 to capture conditionally exposing the debug port.

util/equality/equality_test.go
```diff
@@ -136,3 +141,106 @@ func TestJobConfigChanged(t *testing.T) {
...
+			},
+			expect: true,
+		},
+		{
+			description: "if container args are changed",
+			mutate: func(deploy *appsv1.Deployment) {
+				deploy.Spec.Template.Spec.Containers[0].Args = []string{"foo", "bar", "baz"}
+			},
+			expect: true,
+		},
```
Miciah, line 208: Can you add a test case like this?  https://github.com/openshift/cluster-ingress-operator/blob/c6b66d029f59e71b8d5e345c39beaa819a54d204/pkg/operator/controller/ingress/deployment_test.go#L846-L861

The reason I ask for that test case is kind of subtle, so I'll try to explain:  `DesiredDeployment` defines probes like this:
```go
		LivenessProbe: &corev1.Probe{
			Handler: corev1.Handler{
				HTTPGet: &corev1.HTTPGetAction{
					Path: "/healthz",
					Port: intstr.IntOrString{IntVal: int32(8000)},
				},
			},
		},
```
When the operator creates the resource, the API sets default values for the unspecified fields (`TimeoutSeconds`, `PeriodSeconds`, `SuccessThreshold`, and `FailureThreshold`).

Later the operator gets the deployment and checks the desired and actual deployments.  The API has set values for `TimeoutSeconds` etc., and these values aren't in the deployment that `DesiredDeployment` builds, so `DeploymentConfigChanged` returns true, which causes `updateDeploymentIfNeeded` to try to update to the desired deployment, every time the reconciliation loop runs.  At least this is what I suspect will happen in contour-operator because we have had similar things happen in other operators.  Possible solutions are (a) to define the comparison to treat omitted values the same as the corresponding default values or (b) to define explicit values for any field for which the API sets a default value.

danehans: @Miciah that for catching this. I missed it in my testing, but I'll push a PR to fix the issue by setting the defaults in the desired `Deployment` resource.

stevesloka (approved):
/lgtm

danehans:
Commit `d2bcf00` adds a label/labelSelector that identifies pods of the `Deployment` as being owned by the `Deployment`. The `Deployment` continues to be labeled as being owned by the `Contour`.

controller/contour/deployment.go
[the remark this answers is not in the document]
jpeach: This definitely should be anti-affinity, but we can fix it in a follow-up PR.

jpeach (approved):
LGTM. We can address remaining comments in followup PRs.

PR #63, opened 2020-10-21
Updates ConfigMap Name

danehans:
Previously, the `ConfigMap` name was derived from `contour.Name` . This PR aligns the `ConfigMap` name with upstream until multiple `Contour` instances per namespace are [supported](https://github.com/projectcontour/contour-operator/issues/18).

/assign @jpeach @Miciah 

Signed-off-by: Daneyon Hansen <daneyonhansen@gmail.com>

PR #65, opened 2020-10-21
Updates Deployment Name

danehans:
Previously, the `Deployment` name was derived from `contour.Name + "contour"` . This PR does the following:

- Aligns the `Deployment` name with upstream until multiple `Contour` instances per namespace are [supported](https://github.com/projectcontour/contour-operator/issues/18).
- Updates container port definitions and other port references to use port variables.
- Updates RBAC variable naming
- Fixes equality test to reference both liveness and readiness probes from the correct container name (shutdown-manager).

Requires https://github.com/projectcontour/contour-operator/pull/63 for `contourCfgMapName` variable.

/assign @jpeach @Miciah 
/cc @stevesloka @skriss 

### DIFF
--- a/controller/contour/deployment.go
+++ b/controller/contour/deployment.go
@@ -32,10 +32,12 @@ import (
 )
 
 const (
+	// contourDeploymentName is the name of Contour's Deployment resource.
+	// [TODO] danehans: Remove and use contour.Name + "-contour" when
+	// https://github.com/projectcontour/contour/issues/2122 is fixed.
+	contourDeploymentName = "contour"
 	// contourContainerName is the name of the Contour container.
 	contourContainerName = "contour"
-	// contourCfgMapName is the name of the Contour configmap.
-	contourCfgMapName = "contour"
 	// contourNsEnvVar is the name of the contour namespace environment variable.
 	contourNsEnvVar = "CONTOUR_NAMESPACE"
 	// contourPodEnvVar is the name of the contour pod name environment variable.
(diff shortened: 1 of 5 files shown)


controller/contour/deployment.go
```diff
@@ -32,10 +32,12 @@ import (
...
 
 const (
+	// contourDeploymentName is the name of Contour's Deployment resource.
+	// [TODO] danehans: Remove and use contour.Name + "-contour" when
+	// https://github.com/projectcontour/contour/issues/2122 is fixed.
+	contourDeploymentName = "contour"
 	// contourContainerName is the name of the Contour container.
 	contourContainerName = "contour"
-	// contourCfgMapName is the name of the Contour configmap.
-	contourCfgMapName = "contour"
```
jpeach, line 38: Needs to be on top of the ConfigMap change:
```
controller/contour/deployment.go:290:17: undefined: contourCfgMapName
```

controller/contour/deployment.go
```diff
@@ -55,6 +57,17 @@ const (
 	// contourDeploymentLabel identifies a deployment as a contour deployment,
 	// and the value is the name of the owning contour.
 	contourDeploymentLabel = "contour.operator.projectcontour.io/deployment-contour"
+	// contourDeploymentSvcAcct is the name of the service account
+	// used by the Contour deployment.
+	contourDeploymentSvcAcct = "contour"
```
Miciah, line 62: We should use `contourDeploymentSvcAcct` in `ensureRBAC` as well, but that could be a follow-up.

controller/contour/deployment.go
```diff
@@ -55,6 +57,17 @@ const (
...
+	// xdsPort is the network port number of XDS.
```
Miciah, line 63:
```suggestion
	// xdsPort is the network port number of xDS.
```

jpeach (approved):

PR #83, opened 2020-10-27
Add a multiarch image build for `main`

youngnick:
This commit adds a multiarch image build for the `main` branch only, with
the same mechanisms as mainline Contour. This should give us am64 and arm64
images.

Signed-off-by: Nick Young <ynick@vmware.com>


stevesloka (approved):
/lgtm

Issue #60, opened 2020-10-21
Implement Contour Status Sub Resource


danehans:
**Please describe the problem you have**
Status of `Contour` is currently not implemented. Implement a basic level of status that surfaces an available condition based upon the availability of Contour/Envoy pods, etc..

__Note:__ Other conditions should be implemented post v1alpha1.

xref: [HTTProxy Status Conditions (Detailed)](https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/detailedconditions.go)

/cc @jpeach @Miciah

Miciah:
Example output:
```
$ kubectl get contour/foo -n projectcontour -o yaml
metadata:
  namespace: projectcontour
  name: foo
spec:
  replicas: 2
...
status:
  contourReplicasAvailable: 2
  envoyReplicasAvailable: 6
  certificatesGenerated: "true"
  ...
```

PR #91, opened 2020-10-29
Adds Status Support

danehans:
Adds Contour status support.

Fixes: https://github.com/projectcontour/contour-operator/issues/60

/cc @Miciah @jpeach 


controller/contour/daemonset.go
```diff
@@ -61,25 +61,33 @@ const (
...
+			}
+			return updated, nil
 		}
-		return fmt.Errorf("failed to get daemonset for contour %s/%s: %w", contour.Namespace, contour.Name, err)
+		return nil, fmt.Errorf("failed to get daemonset for contour %s/%s: %w", contour.Namespace, contour.Name, err)
 	}
 
-	if err := r.updateDaemonSetIfNeeded(ctx, current, desired); err != nil {
-		return fmt.Errorf("failed to update daemonset for contour %s/%s: %w", contour.Namespace, contour.Name, err)
+	updated, err := r.updateDaemonSetIfNeeded(ctx, current, desired)
```
sgreene570, line 85: If the daemonset is actually updated here, should `ensureDaemonSet` call `currentDaemonSet` to return the resource living in the cluster _after_ the update?

controller/contour/deployment.go
```diff
@@ -70,25 +70,30 @@ const (
...
 		}
-		return fmt.Errorf("failed to get deployment %s/%s: %w", desired.Namespace, desired.Name, err)
+		return nil, fmt.Errorf("failed to get deployment %s/%s: %w", desired.Namespace, desired.Name, err)
 	}
 
-	if err := r.updateDeploymentIfNeeded(ctx, current, desired); err != nil {
-		return fmt.Errorf("failed to update deployment %s/%s: %w", desired.Namespace, desired.Name, err)
+	updated, err := r.updateDeploymentIfNeeded(ctx, current, desired)
+	if err != nil {
+		return nil, fmt.Errorf("failed to update deployment %s/%s: %w", desired.Namespace, desired.Name, err)
```
sgreene570, line 93: Same comment for `ensureDaemonset`. Should `ensureDeployment` call `currentDeployment` after performing an update?

controller/contour/secret.go
```diff
@@ -0,0 +1,41 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// tlsSecretsAsNsNames returns a list of TLS secrets for the provided
+// contour represented as namespace/secret name.
+func (r *Reconciler) tlsSecretsAsNsNames(ctx context.Context, contour *operatorv1alpha1.Contour) ([]string, error) {
+	current := &corev1.SecretList{}
+	if err := r.Client.List(ctx, current, client.InNamespace(contour.Spec.Namespace.Name)); err != nil {
+		return nil, fmt.Errorf("failed to list current for contour %s/%s", contour.Namespace, contour.Name)
```
sgreene570, line 31: "failed to list current for contour" ->> "failed to list current secret for contour"?

controller/contour/daemonset.go
danehans: `updateDaemonSetIfNeeded()` updates the ds on the api server if `current` doesn't match `desired`. If `updateDaemonSetIfNeeded()` returns an error value, reconciliation of the `Contour` will be triggered and the reconciler will get the latest `Contour` from the api server. Otherwise `updateDaemonSetIfNeeded()` returns a nil error value and `ensureDaemonSet` completes successfully.

controller/contour/deployment.go
danehans: Same feedback as https://github.com/projectcontour/contour-operator/pull/91/files#r518240871.

api/v1alpha1/contour_types.go
```diff
@@ -83,11 +87,37 @@ type NamespaceSpec struct {
...
+	// availableReplicas is the number of observed available Contour
+	// replicas according to the deployment. The deployment and its
+	// pods will reside in the namespace specified in spec.namespace.name
+	// of the contour.
 	AvailableReplicas int32 `json:"availableReplicas"`
+
+	// availablePods is the number of observed available Envoy pods.
+	// The daemonset and its pods will reside in the namespace specified
+	// in spec.namespace.name of the contour.
+	AvailablePods int32 `json:"availablePods"`
```
skriss, line 106: Personally I do think it'd be clearer if these two field names differentiated between Contour pods and Envoy pods - e.g. `AvailableContourPods` and `AvailableEnvoyPods` or something like that.

controller/contour/controller.go
```diff
@@ -149,6 +151,9 @@ func (r *Reconciler) ensureContour(ctx context.Context, contour *operatorv1alpha
 	if err := r.ensureEnvoyService(ctx, contour); err != nil {
 		return fmt.Errorf("failed to ensure service for contour %s/%s: %w", contour.Namespace, contour.Name, err)
 	}
+	if err := r.syncContourStatus(ctx, contour, deploy, ds); err != nil {
```
skriss, line 154: does it matter that if one of the `ensureXXX` functions returns an error above so `ensureContour` returns early, that no status will be created? I wonder if it'd be better/possible to update it as you go, or always update it before returning.

controller/contour/secret.go
danehans: This pkg has been removed.

danehans:
Commit `26d3ab9` addresses the feedback from @skriss.

api/v1alpha1/contour_types.go
```diff
@@ -83,11 +87,36 @@ type NamespaceSpec struct {
...
 // ContourStatus defines the observed state of Contour.
 type ContourStatus struct {
-	// availableReplicas is the number of observed available Contour replicas
-	// according to the deployment.
-	AvailableReplicas int32 `json:"availableReplicas"`
+	// availableContours is the number of observed available replicas
+	// according to the Contour deployment. The deployment and its pods
+	// will reside in the namespace specified by spec.namespace.name of
+	// the contour.
+	AvailableContours int32 `json:"AvailableContours"`
```
stevesloka, line 101: I like the rename here! 👍 

We should think about future stats like "requestedContours" or something to contrast how many are healthy vs requested.

controller/contour/status_test.go
```diff
@@ -0,0 +1,250 @@
+package contour
```
stevesloka, line 1: Needs license header.

util/equality/equality.go
```diff
@@ -209,3 +211,21 @@ func LoadBalancerServiceChanged(current, expected *corev1.Service) (*corev1.Serv
 
 	return updated, true
 }
+
+// ContourStatusChanged checks if current and expected match and if not,
+// returns true.
+func ContourStatusChanged(current, expected operatorv1alpha1.ContourStatus) bool {
```
stevesloka, line 217: Could you cover this with some tests just to outline the desired functionality?

stevesloka (changes requested):

danehans:
@stevesloka commit `0a34ff2` updates the PR based on your comments.

api/v1alpha1/contour_types.go
skriss, line 101:
```suggestion
	AvailableContours int32 `json:"availableContours"`
```

danehans:
@skriss commit `6fc3d2b` addresses your comments in https://github.com/projectcontour/contour-operator/pull/91#discussion_r520153008.

api/v1alpha1/contour_types.go
jpeach: Requested Contour replicas would be in the spec, right?

stevesloka:
@danehans I think this is ready to go, everyone's comments have been addressed it seems, just have a conflict on the crds it looks like.

skriss (approved):
no further comments from me, LGTM once conflict is resolved

cmd/contour-operator.go
```diff
@@ -60,29 +61,24 @@ func main() {
...
 
-	if err = (&contourcontroller.Reconciler{
-		Config: contourcontroller.Config{
-			ContourImage: contourImage,
-			EnvoyImage:   envoyImage,
-		},
-		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("Contour"),
-		Scheme: mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
```
Miciah, line 82: Should we update `controller/contour/suite_test.go` similarly? (Follow-up?)

controller/contour/controller.go
```diff
@@ -119,71 +121,88 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		return fmt.Errorf("failed to finalize contour %s/%s: %w", contour.Namespace, contour.Name, err)
+		errs = append(errs, fmt.Errorf("failed to finalize contour %s/%s: %w", contour.Namespace, contour.Name, err))
 	}
 	if err := r.ensureNamespace(ctx, contour); err != nil {
-		return fmt.Errorf("failed to ensure namespace %s for contour %s/%s: %w",
-			contour.Spec.Namespace.Name, contour.Namespace, contour.Name, err)
+		errs = append(errs, fmt.Errorf("failed to ensure namespace %s for contour %s/%s: %w",
+			contour.Spec.Namespace.Name, contour.Namespace, contour.Name, err))
 	}
 	if err := r.ensureRBAC(ctx, contour); err != nil {
```
Miciah, line 134: Is it useful to call `ensureRBAC` if there was an error while ensuring the namespace?  Some of the resources that `ensureRBAC` creates are namespaced, and the deployment and daemonset are namespaced, so calling `ensureRBAC` if the namespace does not exist generates additional errors and cannot produce a running Contour or Envoy instance.

controller/contour/controller.go
```diff
@@ -119,71 +121,88 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	deploy := &appsv1.Deployment{}
+	ds := &appsv1.DaemonSet{}
+	if len(errs) == 0 {
+		if err := r.ensureConfigMap(ctx, contour); err != nil {
+			errs = append(errs, fmt.Errorf("failed to ensure configmap for contour %s/%s: %w", contour.Namespace, contour.Name, err))
+		}
+		if err := r.ensureJob(ctx, contour); err != nil {
+			errs = append(errs, fmt.Errorf("failed to ensure job for contour %s/%s: %w", contour.Namespace, contour.Name, err))
+		}
+		deploy, err = r.ensureDeployment(ctx, contour)
```
Miciah, line 148: Can we move `err` into the inner scope and to the point of use?  
```suggestion
	deploy := &appsv1.Deployment{}
	ds := &appsv1.DaemonSet{}
	if len(errs) == 0 {
		if err := r.ensureConfigMap(ctx, contour); err != nil {
			errs = append(errs, fmt.Errorf("failed to ensure configmap for contour %s/%s: %w", contour.Namespace, contour.Name, err))
		}
		if err := r.ensureJob(ctx, contour); err != nil {
			errs = append(errs, fmt.Errorf("failed to ensure job for contour %s/%s: %w", contour.Namespace, contour.Name, err))
		}
		var err error
		deploy, err = r.ensureDeployment(ctx, contour)
```

controller/contour/status.go
```diff
@@ -0,0 +1,164 @@
...
+
+	return retryable.NewMaybeRetryableAggregate(errs)
+}
+
+// computeContourAvailableCondition computes the contour Available status condition
+// type based on deployment and ds. The contour is only available if the deployment's
+// "Available" condition is "True" and the status numberAvailable of ds greater than
+// or equal to 1
+func computeContourAvailableCondition(deployment *appsv1.Deployment, ds *appsv1.DaemonSet) metav1.Condition {
+	if deployment == nil {
```
Miciah, line 58: `deployment` cannot be nil or `syncContourStatus` would have panicked on `updated.Status.AvailableContours = deployment.Status.AvailableReplicas`.

(If you do make `syncContourStatus` handle nil values for `deployment` and `ds`, then you could simplify `ensureContour` slightly by declaring the `deploy` and `ds` variables without initializing them to pointers to zero values.)

controller/contour/status.go
```diff
@@ -0,0 +1,164 @@
...
+		return metav1.Condition{
+			Type:    operatorv1alpha1.ContourAvailableConditionType,
+			Status:  metav1.ConditionFalse,
+			Reason:  "ContourUnavailable",
+			Message: "Contour deployment does not exist.",
+		}
+	}
+
+	if ds == nil {
```
Miciah, line 67: `ds` cannot be nil.

controller/contour/status.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	dsAvailable := ds.Status.NumberAvailable > 0
+	for _, cond := range deployment.Status.Conditions {
+		if cond.Type != appsv1.DeploymentAvailable {
+			continue
+		}
+		switch {
+		case cond.Status == corev1.ConditionTrue:
+			if !dsAvailable {
```
Miciah, line 83: This logic is kind of confusing.  What do you think about the following alternative procedure?

1. Compute an `EnvoyAvailable` condition using `dsAvailable`.
2. Compute a `ContourAvailable` condition using `deployment.StatusConditions`.
3. Compute a general `Available` condition using the `EnvoyAvailable` and `ContourAvailable` conditions.

Maybe in a follow-up?

controller/contour/status.go
```diff
@@ -0,0 +1,164 @@
...
+	for i, update := range updates {
+		add := true
+		for j, cond := range conditions {
+			if cond.Type == update.Type {
+				add = false
+				if conditionChanged(cond, update) {
+					conditions[j].Status = update.Status
+					conditions[j].Reason = update.Reason
+					conditions[j].Message = update.Message
+					conditions[j].LastTransitionTime = now
```
Miciah, line 148: "lastTransitionTime is the last time the condition transitioned from one status to another."  So we should only update `conditions[j].LastTransitionTime` if `cond.Status != update.Status`.  (We have arguably been doing it incorrectly in other operators.)

https://github.com/kubernetes/apimachinery/blob/000b5f4f8623fa1a54269300e8987dfbf70033ed/pkg/apis/meta/v1/types.go#L1390-L1391

controller/contour/status_test.go
```diff
@@ -0,0 +1,263 @@
...
+	testCases := []struct {
+		description      string
+		deployConditions []appsv1.DeploymentCondition
+		dsAvailable      int32
+		expect           metav1.Condition
+	}{
+		{
+			description: "deployment progressing, daemonset unavailable",
+			deployConditions: []appsv1.DeploymentCondition{
+				{Type: appsv1.DeploymentProgressing, Status: corev1.ConditionTrue},
```
Miciah, line 50: `computeContourAvailableCondition` only looks at `appsv1.DeploymentAvailable`, so maybe the description should be "deployment has no Available condition".

controller/contour/status_test.go
```diff
@@ -0,0 +1,263 @@
...
+			dsAvailable: int32(0),
+			expect: metav1.Condition{
+				Type:   operatorv1alpha1.ContourAvailableConditionType,
+				Status: metav1.ConditionUnknown,
+			},
+		},
+		{
+			description: "deployment failure, daemonset unavailable",
+			deployConditions: []appsv1.DeploymentCondition{
+				{Type: appsv1.DeploymentReplicaFailure, Status: corev1.ConditionTrue},
```
Miciah, line 61: This is effectively the same as the previous test case, isn't it?

controller/contour/status_test.go
```diff
@@ -0,0 +1,263 @@
...
+			a:           metav1.Condition{},
+		},
+		{
+			description: "empty slices should be equal",
+			expected:    false,
+			a:           metav1.Condition{},
+			b:           metav1.Condition{},
+		},
+		{
+			description: "condition LastTransitionTime should not be ignored",
```
Miciah, line 149:
```suggestion
			description: "condition LastTransitionTime should be ignored",
```

controller/contour/status_test.go
```diff
@@ -0,0 +1,263 @@
...
+		{
+			description: "reason updated",
+			current: []metav1.Condition{
+				newCondition("available", "false", "Reason", "Message", start),
+			},
+			updates: []metav1.Condition{
+				newCondition("available", "false", "New Reason", "Message", middle),
+			},
+			expected: []metav1.Condition{
+				newCondition("available", "false", "New Reason", "Message", later),
```
Miciah, line 236:
```suggestion
				newCondition("available", "false", "New Reason", "Message", start),
```

controller/contour/status_test.go
```diff
@@ -0,0 +1,263 @@
...
+		{
+			description: "message updated",
+			current: []metav1.Condition{
+				newCondition("available", "false", "Reason", "Message", start),
+			},
+			updates: []metav1.Condition{
+				newCondition("available", "false", "Reason", "New Message", middle),
+			},
+			expected: []metav1.Condition{
+				newCondition("available", "false", "Reason", "New Message", later),
```
Miciah, line 248:
```suggestion
				newCondition("available", "false", "Reason", "New Message", start),
```

controller/manager/manager.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, fmt.Errorf("failed to create contour manager: %w", err)
+	}
+
+	reconciler := newContourReconciler(mgr, cfg)
+
+	if err = builder.
+		ControllerManagedBy(mgr).
+		For(&operatorv1alpha1.Contour{}).
+		Watches(&source.Kind{Type: &appsv1.Deployment{}}, enqueueRequestForOwningContour()).
+		Watches(&source.Kind{Type: &appsv1.DaemonSet{}}, enqueueRequestForOwningContour()).
```
Miciah, line 59: We didn't have these watches before, right?  These watches mean not only that the operator updates the CR's status in response to updates to the operands, but also that the operator reconciles and reverts changes to the operands by external agents.  It is a little surprising to see in a change related to status, so I wanted to call it out, but it is desirable behavior, so might as well go ahead and do it now.

Miciah (commented):
A few comments, but nothing that couldn't be deferred to follow-ups.

controller/manager/manager.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+
+	operatorv1alpha1 "github.com/projectcontour/contour-operator/api/v1alpha1"
+	contourcontroller "github.com/projectcontour/contour-operator/controller/contour"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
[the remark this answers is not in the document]
danehans: It was included in the [kubebuilder bootstrap](https://book.kubebuilder.io/cronjob-tutorial/empty-main.html). I simply moved it to this manager pkg.

util/equality/equality_test.go
```diff
@@ -559,3 +560,82 @@ func TestLoadBalancerServiceChanged(t *testing.T) {
...
+					{
+						Type:    "Available",
+						Status:  metav1.ConditionFalse,
+						Reason:  "Foo",
+						Message: "Changed",
+					},
+				}
+			},
+			expect: true,
+		},
```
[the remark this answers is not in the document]
danehans: Since we only have one condition, I was planning to add ^ test case when additional conditions are introduced. I'll leave a TODO in the test code as a future reminder.

cmd/contour-operator.go
danehans: @Miciah I tried, but I get an import cycle error:
```
$ make test
/Users/daneyonhansen/code/go/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
controller/contour/suite_test.go
go vet ./...
package github.com/projectcontour/contour-operator/controller/contour (test)
	imports github.com/projectcontour/contour-operator/controller/manager
	imports github.com/projectcontour/contour-operator/controller/contour: import cycle not allowed in test
make: *** [vet] Error 1
```

See [my branch](https://github.com/danehans/contour-operator/tree/suite_test_refactor_mgr_pkg) for details.

controller/contour/controller.go
danehans: Understood. I took this approach so `syncContourStatus()` is called without requiring conditionals for the ensure namespace and rbac funcs. Any suggestions?

controller/contour/status.go
danehans: I created https://github.com/projectcontour/contour-operator/issues/113 to track this as a follow-up.

cmd/contour-operator.go
danehans: I created https://github.com/projectcontour/contour-operator/issues/114 to track this issue as a follow-up.

controller/contour/status_test.go
danehans: The previous is available unknown and this is a different condition type. Is it unnecessary to test for both to ensure the code behaves as expected?

controller/manager/manager.go
danehans: > We didn't have these watches before, right?

That's correct.

> It is a little surprising to see in a change related to status, so I wanted to call it out, but it is desirable behavior, so might as well go ahead and do it now.

I thought the controller needs these watches so it can update the status of `Contour`. In my testing, the status of `Contour` does not get updated without these watches.

danehans:
@Miciah I have addressed your feedback either inline or in commit `bec8cde`, PTAL.

controller/contour/controller.go
Miciah: It would make sense to me to use conditionals.  That is, `ensureNamespace` could return a Boolean value indicating whether or not the namespace exists, and the call to `ensureRBAC` could be wrapped in an `if haveNamespace {` block.  (This is how I approach "ensure" methods in general:  If resource Y requires resource X, then `ensureX` should return a Boolean return value `haveX`, and the call to `ensureY` should be wrapped in `if haveX {`.)

controller/contour/status_test.go
Miciah: It looks a little odd since the logic being tested ignores both conditions, but I don't really mind if you prefer to keep these test cases as they are.

controller/manager/manager.go
Miciah: > > It is a little surprising to see in a change related to status, so I wanted to call it out, but it is desirable behavior, so might as well go ahead and do it now.
> 
> I thought the controller needs these watches so it can update the status of `Contour`. In my testing, the status of `Contour` does not get updated without these watches.

Right, sorry I wasn't more clear.  These changes are necessary for proper status updating (that's the intended main effect).  They have a side effect as well, namely that they cause the operator to reconcile modifications to the resources by others.  The main effect and side effect are both desirable; I just want to call out the side effect as noteworthy.

controller/contour/controller.go
```diff
@@ -128,66 +130,83 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		return fmt.Errorf("failed to ensure deployment for contour %s/%s: %w", contour.Namespace, contour.Name, err)
-	}
-	if err := r.ensureDaemonSet(ctx, contour); err != nil {
-		return fmt.Errorf("failed to ensure daemonset for contour %s/%s: %w", contour.Namespace, contour.Name, err)
-	}
-	if err := r.ensureContourService(ctx, contour); err != nil {
-		return fmt.Errorf("failed to ensure service for contour %s/%s: %w", contour.Namespace, contour.Name, err)
+
+	deploy := &appsv1.Deployment{}
+	ds := &appsv1.DaemonSet{}
```
Miciah, line 143: Tiny suggestion:
```suggestion
	var deploy appsv1.Deployment
	var ds appsv1.DaemonSet
```
now that `syncContourStatus` checks for nil values.

Miciah:
If I look at the changes on GitHub, I see that the linter has a few comments about some instances of `else` after a `return`... but that's just lint.  * grin *.

controller/contour/status_test.go
danehans: @Miciah I'm trying to verify that the fall through case gets hit when the deployment condition is `ReplicaFailure` or `Progressing`:

```
	return metav1.Condition{
		Type:    operatorv1alpha1.ContourAvailableConditionType,
		Status:  metav1.ConditionUnknown,
		Reason:  "ContourUnknown",
		Message: "Contour status unknown.",
	}
```

Let me know if you prefer I explicitly state this in core or if you prefer I remove the test cases.

controller/contour/controller.go
danehans: Changing the vars as you suggest are an issue with `ensureDeployment(ctx, contour)` and `ensureDaemonSet(ctx, contour)`. Both funcs return pointers for these vars instead of values.

controller/manager/manager.go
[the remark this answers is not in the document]
danehans: I removed `_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"` since the linter complained:
```
controller/manager/manager.go:26:2: a blank import should be only in a main or test package, or have a comment justifying it (golint)
	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```

danehans:
> If I look at the changes on GitHub, I see that the linter has a few comments about some instances of else after a return... but that's just lint. * grin *.

Yeah, I opened https://github.com/projectcontour/contour-operator/issues/115 to add the ability to run the linter locally. Commit `837fda5` addressed the linter issue and the PR is now passing CI.

controller/contour/controller.go
danehans: @Miciah should I update these func to return values instead of pointers?

Miciah: > Changing the vars as you suggest are an issue with `ensureDeployment(ctx, contour)` and `ensureDaemonSet(ctx, contour)`. Both funcs return pointers for these vars instead of values.

Ah, sorry, I meant this:
```go

> @Miciah should I update these func to return values instead of pointers?

**Comment by Miciah** [comment]
/lgtm
We can handle any remaining minor issues with follow-ups if necessary.

File: api/v1alpha1/contour_types.go [hunk 25]
```diff
@@ -83,11 +87,36 @@ type NamespaceSpec struct {
...
+	// Available indicates that the contour is running and available.
+	ContourAvailableConditionType = "Available"
+)
+
 // ContourStatus defines the observed state of Contour.
 type ContourStatus struct {
-	// availableReplicas is the number of observed available Contour replicas
-	// according to the deployment.
-	AvailableReplicas int32 `json:"availableReplicas"`
+	// availableContours is the number of observed available replicas
```
stevesloka, line 97: nit: s/availableContours/AvailableContours

api/v1alpha1/contour_types.go
```diff
@@ -83,11 +87,36 @@ type NamespaceSpec struct {
...
+	// according to the Contour deployment. The deployment and its pods
+	// will reside in the namespace specified by spec.namespace.name of
+	// the contour.
+	AvailableContours int32 `json:"availableContours"`
+
+	// availableEnvoys is the number of observed available pods from
```
stevesloka, line 103: nit: s/availableEnvoys/AvailableEnvoys

api/v1alpha1/contour_types.go
```diff
@@ -83,11 +87,36 @@ type NamespaceSpec struct {
...
+	// the Envoy daemonset. The daemonset and its pods will reside in the
+	// namespace specified by spec.namespace.name of the contour.
+	AvailableEnvoys int32 `json:"availableEnvoys"`
+
+	// conditions represent the observations of a contour's current state.
```
stevesloka, line 108: nit: s/conditions/Conditions

stevesloka (approved):
/lgtm with just some small blah nits. =)

danehans:
@stevesloka commit `2d43a78` lowercases the field names introduced by this PR. I'll submit a separate PR to uppercase other field names in the API.

PR #93, opened 2020-10-30
makefile: Add golangci-lint make task

### LINKED ISSUE #92: makefile: Add golanglint-ci to tasks
Mirroring what Contour does, we should add the golanglint-ci checks at least as a starting point to help keep code consistent.

stevesloka:
Adds a `lint-golint` task in the Makefile to run golangci-lint on the codebase. Also adds `check` task to run tests & the linting in one step while developing.

The second commit in the PR fixes up the errors found from running the linting task. 

@danehans there are a bunch of methods that aren't used that I hated to just delete, so they are commented out right now. If we don't need them then we should just remove. =)
    
Closes #92

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,7 @@
 bin/
 
 # Dependency directories (remove the comment below to include it)
-# vendor/
+vendor/
 
 # Goland
 .idea/
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,43 @@
+run:
+  timeout: 10m
+
+linters:
+  enable:
+    - bodyclose
+    - gofmt
+    - goimports
+    - golint
+    - gosec
+    - misspell
+    - unconvert
+    - unparam
+
+linters-settings:
+  misspell:
+    ignore-words:
+      - clas
+      - cancelled
+    locale: US
+  gofmt:
+    simplify: true
+  unparam:
+    check-exported: false
+
+issues:
+  exclude:
+    # TODO(jpeach): exclude unparam warnings about functions that always receive
+    # the same arguments. We should clean those up some time.
+    - always receives
+  exclude-rules:
+    - path: zz_generated
+      linters:
+        - goimports
+    # TODO(youngnick): Disable the deprecation warnings, because of the "github.com/golang/protobuf/proto"
+    # libraries. We can't stop using them yet, because they move from Proto v1 to V2
+    # And we can't nolint them by line, because https://github.com/golangci/golangci-lint/issues/741
+    - linters:
+        - golint
+      text: "returns unexported type"
+    - linters:
+        - golint
+      text: "don't use ALL_CAPS in Go names; use CamelCase"
--- a/Makefile
+++ b/Makefile
@@ -64,9 +64,17 @@ endif
 
 all: manager
 
+# Run tests & validate against linters
+.PHONY: check
+check: test lint-golint
+
 # Run tests
 test: generate fmt vet manifests
 	go test ./... -coverprofile cover.out
+
+lint-golint:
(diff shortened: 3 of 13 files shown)


skriss (commented):
Mostly LGTM. Re: the unused functions - it *looks* like that functionality is implemented in https://github.com/projectcontour/contour-operator/blob/main/controller/contour/rbac.go#L269, so the unused helpers can actually be deleted, but let's get confirmation from @danehans.

stevesloka:
@danehans any thoughts on the unused methods noted in the PR?

.gitignore
```diff
@@ -15,7 +15,7 @@
 bin/
 
 # Dependency directories (remove the comment below to include it)
-# vendor/
+vendor/
 
 # Goland
-.idea/
+.idea/
```
danehans, line 21: Newline needed

.golangci.yml
```diff
@@ -0,0 +1,52 @@
...
+    # And we can't nolint them by line, because https://github.com/golangci/golangci-lint/issues/741
+    - linters:
+        - staticcheck
+      text: "SA1019: package github.com/golang/protobuf"
+    - linters:
+        - golint
+      text: "var `ingress_https` should be `ingressHTTPS`"
+    - linters:
+        - golint
+      text: "var `ingress_http` should be `ingressHTTP`"
```
danehans, line 46: Are these applicable to the operator?

danehans:
@stevesloka I'll tag this PR after addressing my comments above and rebasing on https://github.com/projectcontour/contour-operator/pull/93.

.golangci.yml
stevesloka: nope, I just copy/pasted from Contour, I removed them.

.gitignore
stevesloka: 👍

stevesloka:
all fixed up @danehans, please take a look.

Makefile
```diff
@@ -64,10 +64,17 @@ endif
 
 all: manager
 
+.PHONY: checkall
```
danehans, line 67: s/checkall/check/

Makefile
```diff
@@ -64,10 +64,17 @@ endif
...
+check: test lint-golint
```
danehans, line 68: Since the current target flow is `all` > `manager` > `test`, I suggest removing the `test` target from `check` and replace `lint-golint` with `check` in L71.

danehans:
@stevesloka I had a few nits, otherwise LGTM. I would also appreciate if @jpeach or @skriss can provide a review.

util/equality/equality.go
```diff
@@ -114,7 +114,7 @@ func DeploymentConfigChanged(current, expected *appsv1.Deployment) (*appsv1.Depl
 // ClusterIpServiceChanged checks if the spec of current and expected match and if not,
```
skriss, line 114:
```suggestion
// ClusterIPServiceChanged checks if the spec of current and expected match and if not,
```

skriss (approved):
LGTM apart from open comments, will be nice to get the `main` build to green!

Makefile
jpeach: "check" and "test" should be the same. "check" is for people who have muscle memory from auto tools :)

jpeach (approved):
LGTM once remaining issues are resolved.

Makefile
stevesloka: > "check" and "test" should be the same.

I disagree. "test" should run your tests, "check" should run your tests and verify other checks against the branch.

stevesloka: @danehans I took out the linting from `test` which should keep `test` the way it was before. To run linters then you need to run `check` which will do both.

PR #100, opened 2020-11-04
Refactors Controller Reconciler

danehans:
Previously, `ensureContour()` was called twice for every reconciliation of a new `Contour`. This is because appending a finalizer modifies the object causing a reconciliation. This PR makes `ensureContour()` conditional based on the presence of the finalizer.

/assign @jpeach @stevesloka @skriss 
/cc @Miciah 

### DIFF
--- a/controller/contour/controller.go
+++ b/controller/contour/controller.go
@@ -87,12 +87,21 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// The contour is safe to process, so ensure current state matches desired state.
 	desired := contour.ObjectMeta.DeletionTimestamp.IsZero()
 	if desired {
-		if err := r.ensureContour(ctx, contour); err != nil {
-			switch e := err.(type) {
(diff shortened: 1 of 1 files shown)


controller/contour/controller.go
```diff
@@ -247,3 +251,16 @@ func envoyDaemonSetPodSelector(contour *operatorv1alpha1.Contour) *metav1.LabelS
 		},
 	}
 }
+
+// contourFinalized returns true if contour is finalized.
+func contourFinalized(contour *operatorv1alpha1.Contour) bool {
+	if contour.Finalizers != nil {
```
skriss, line 257: nit: you shouldn't need to nil-check here, since range over a nil slice won't panic

skriss (approved):
LGTM, one nit

danehans:
@skriss commit `0dc93a1` addresses your feedback in https://github.com/projectcontour/contour-operator/pull/100#discussion_r519980395.

skriss (approved):

PR #102, opened 2020-11-05
Updates Operator RBAC Rules

### LINKED ISSUE #78: Remove endpoint RBAC
**What steps did you take and what happened:**

The operator doesn't manage endpoints so we can remove those from the RBAC config.

danehans:
Updates operator RBAC to allow get|list|watch verbs for endpoints and all verbs for services. __Note:__ The operator requires all verbs for services since it manages contour/envoy services. Get|List|Watch verbs for endpoints is required so the operator can delegate to the [contour ClusterRole](https://github.com/projectcontour/contour/blob/main/examples/contour/02-role-contour.yaml#L20-L27). 

Closes #78 

/assign @jpeach @stevesloka 
/cc @Miciah 

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -21,10 +21,7 @@ rules:
   - ""
   resources:
   - endpoints
-  - services
   verbs:
-  - create
-  - delete
   - get
   - list
   - watch
@@ -34,6 +31,7 @@ rules:
   - namespaces
   - secrets
   - serviceaccounts
+  - services
   verbs:
   - create
   - delete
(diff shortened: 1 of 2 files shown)


config/rbac/role.yaml
```diff
@@ -41,6 +38,17 @@ rules:
   - list
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - services
```
skriss, line 44: nit: looks like you could move `services` into the rule just above with `namespaces`, etc., if you want to keep similar things together.

skriss (approved):
LGTM, just an optional nit.

jpeach (approved):
LGTM.

FWIW, my preference would be for each `
+kubebuilder:rbac` to reference a single resource. I think that's easier to reason about.

I'll leave the merge to @skriss

danehans:
@skriss commit `05e4570` addresses your feedback in https://github.com/projectcontour/contour-operator/pull/102#discussion_r518408953

skriss:
> FWIW, my preference would be for each ` +kubebuilder:rbac` to reference a single resource. I think that's easier to reason about.

Fair enough, that would work for me too.

PR #103, opened 2020-11-05
Updates Makefile Docker Targets

danehans:
#83 added docker-related make targets. Similar targets existed from kubebuilder bootstrapping. This PR removes docker targets created from kubebuilder bootstrapping and calls the `test` target before building the docker image.

/assign @stevesloka @jpeach 
/cc @youngnick @Miciah


skriss (approved):
LGTM

Issue #55, opened 2020-10-19
contour/envoy container images should come from CRD


stevesloka:
The image version that is used for Contour/Envoy are taken from an command line arg to the operator, but the operator should be able to derive the information from the Contour-Operator CRD such than different instance of Contour can run their own version requested via the CRD.

_Originally posted by @danehans in https://github.com/projectcontour/contour-operator/pull/44#discussion_r505921582_

Miciah:
In `config/manager/manager.yaml`, we should be using environment variables such as `CONTOUR_IMAGE` and `ENVOY_IMAGE`, and interpolating those into the command like, something like the following:

```yaml
      containers:
      - name: manager
        command:
        - /contour-operator
        args:
        - --enable-leader-election
        - --contour-image
        - "$(CONTOUR_IMAGE)"
        - --envoy-image
        - "$(ENVOY_IMAGE)"
        env:
          - name: CONTOUR_IMAGE
            value: oldX
          - name: ENVOY_IMAGE
            value: oldY
```

And in `kustomization.yaml`, we should replace the values for those variables:

```yaml
spec:
  template:
    spec:
      containers:
        - name: manager
          env:
            - name: CONTOUR_IMAGE
              value: newX
            - name: ENVOY_IMAGE
              value: newY
```

Reference:  https://kubectl.docs.kubernetes.io/pages/reference/kustomize.html

@stevesloka, @skriss, @jpeach, does this approach look reasonable?

stevesloka:
There should be an `image` field in the spec here: https://github.com/projectcontour/contour-operator/blob/main/api/v1alpha1/contour_types.go#L50

That value should be used for each instance of Contour to allow you to define what image of Contour/Endoy you'd like to run. It could be upstream images or custom depending on the user needs.

Miciah:
What is the use-case for upgrading Contour instances independently?  It seems like leaving it up to users to update the CRs could devolve into an unsupportable matrix of different (and possibly incompatible) Contour and Envoy versions.  

One of the goals of the operator should be to handle these details.  In the most common case, a given release of contour-operator will know the corresponding releases of Contour and Envoy that the operator manages and should handle upgrading (or possibly downgrading) to the appropriate images.  As mentioned in #71, we'd like to synchronize projectcontour/contour-operator releases with projectcontour/contour releases.  If the user wants to upgrade Contour, the user should upgrade the operator to the appropriate release.

stevesloka:
[...]

Issue #71, opened 2020-10-23
Incorporate projectcontour/contour-operator in to the Contour release process


Miciah:
We would like to incorporate contour-operator into [the Contour release process](https://projectcontour.io/resources/release-process/).  

Goals:

* Enable us to make a new projectcontour/contour-operator release shortly after, or in tandem with, each new projectcontour/contour release.  
* Keep projectcontour/contour and projectcontour/contour-operator releases in sync as much as possible.  
* Automate the contour-operator release process.  

To this end, we propose the following changes:

* Define a GitHub action that triggers on each new projectcontour/contour release.  This action would open a PR in projectcontour/contour-operator to update `config/manager/kustomization.yaml` to reference the new Contour image.  CI would then automatically verify contour-operator with the new Contour release.  The PR description could include instructions for reviewing the PR and completing the projectcontour/contour-operator release.  

* Update the Contour release process with an additional/follow-up step to do a manual review of the automatically created PR.  If the PR is passing tests with the new Contour release and there are no outstanding issues in projectcontour/contour-operator, a maintainer would merge the PR and follow the instructions in the PR's description to create and push a tag, which would produce a new contour-operator release.  

This would be very nice to have for v1alpha1, but it is more a stretch goal than a blocking issue.  

---

As a long-term goal once contour-operator is more mature, it would make sense to use projectcontour/contour-operator to set up Contour in projectcontour/contour CI jobs, but we can track that as a separate issue.  

---

@jpeach, @stevesloka, does this approach seem feasible to you?

stevesloka:
Sounds good to me @Miciah! At a minumim we can write up the release notes on how to deploy the operator and follow along manually until all the automation bits are up: https://projectcontour.io/resources/release-process/

jpeach:
Maybe we need to think more about this. If users install the operator, then they no longer have to upgrade Contour, but they have to upgrade the operator instead. If upgrading Contour is the same thing as upgrading the operator, then the operator won't necessarily be able to give a good feedback UX on the upgrade process. If this is the approach, then it's hard for me to see how you could do a rollback if an upgrade turns out to be a bad idea. I'm sure that other operators have similar concerns; is there prior art in this area that e could learn from?

danehans:
> If upgrading Contour is the same thing as upgrading the operator, then the operator won't necessarily be able to give a good feedback UX on the upgrade process.

[...]

PR #116, opened 2020-11-16
Adds Support for Release Automation

danehans:
Ports the Contour [final release](https://projectcontour.io/resources/release-process/#final-release) process to Contour Operator.

Fixes: https://github.com/projectcontour/contour-operator/issues/55
xref: https://github.com/projectcontour/contour-operator/issues/71

/assign @stevesloka @skriss @jpeach 
/cc @Miciah 

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -25,9 +25,13 @@ spec:
       containers:
       - command:
         - /contour-operator
+        - --contour-image
+        - docker.io/projectcontour/contour-operator:main
+        - --envoy-image
+        - docker.io/envoyproxy/envoy:v1.16.0
         args:
         - --enable-leader-election
-        image: controller:latest
+        image: docker.io/projectcontour/contour:main
         name: manager
         resources:
           limits:
--- a/hack/release/make-release-tag.sh
+++ b/hack/release/make-release-tag.sh
@@ -1,0 +1,85 @@
+#! /usr/bin/env bash
+
+# make-release-tag.sh: This script assumes that you are on a branch and have
+# otherwise prepared the release. It rewrites the image tag in the operator's
+# deployment manifest, then creates a tag with a message containing the shortlog
+# from the previous version. Note: The Envoy image is not managed by this script.
+
+readonly PROGNAME=$(basename "$0")
+readonly OLDVERS="$1"
+readonly NEWVERS="$2"
+
+if [ -z "$OLDVERS" ] || [ -z "$NEWVERS" ]; then
+    printf "Usage: %s OLDVERS NEWVERS\n" "$PROGNAME"
+    exit 1
+fi
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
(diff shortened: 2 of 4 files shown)


danehans:
__Note:__ This PR does not manage the Envoy image tag. @stevesloka is this tag managed manually?

config/manager/manager.yaml
```diff
@@ -25,9 +25,13 @@ spec:
...
       - command:
         - /contour-operator
+        - --contour-image
+        - docker.io/projectcontour/contour:main
+        - --envoy-image
+        - docker.io/envoyproxy/envoy:v1.16.0
         args:
         - --enable-leader-election
-        image: controller:latest
+        image: docker.io/projectcontour/contour:main
```
stevesloka, line 34: Should be the operator image?

stevesloka:
@danehans yeah the Envoy image is updated when either a major version comes our or a patched version. The patches don't carry Contour changes when only the Envoy image updates (i.e. CVEs in Envoy).

hack/release/make-release-tag.sh
```diff
@@ -0,0 +1,84 @@
...
+    esac
+}
+
+# Update the Docker image tags in the operator's deployment manifest.
+# The version might be main or OLDVERS depending on whether we are
+# tagging from the release branch or from main.
+run::sed \
+    "-es|docker.io/projectcontour/contour:main|$IMG|" \
+    "-es|docker.io/projectcontour/contour:$OLDVERS|$IMG|" \
+    "-es|docker.io/projectcontour/contour-operator:main|$IMG|" \
```
stevesloka, line 47: This needs a new var above since it would replace the `contour-operator` image with `contour` on lines 46 & 47.

stevesloka (changes requested):

hack/release/make-release-tag.sh
danehans: @stevesloka I don't believe so since we plan on synchronizing contour/operator releases. The operator should have the same release version as Contour. For example v1alpha1 of the operator will be v1.11 when released on 12/4 with contour v1.11.

danehans:
@stevesloka commit `279d161` resolves https://github.com/projectcontour/contour-operator/pull/116#discussion_r524472157. I'm waiting for feedback from https://kubernetes.slack.com/archives/C8XRH2R4J/p1605569514076200 before resolving https://github.com/projectcontour/contour-operator/pull/116#discussion_r524474216.

hack/release/make-release-tag.sh
skriss: @danehans looks like the issue is that L47 and L48 replace not just the version/tag for the operator, but replace the entire `contour-operator:<version>` string with `contour:<version>` which is not desirable.

danehans: @skriss commit `001327e` fixes the issue. Thanks for the explanation.

stevesloka (approved):
/lgtm

skriss (approved):
LGTM

Issue #9, opened 2020-09-01
Add linter configuration.


jpeach:
**Please describe the problem you have**

Add a golangci-lint configuration to make it easier for contributors to match Go style. This reduces the number of nits that need to be addressed in code review.

danehans:
Currently, golint-ci runs against PRs as a CI job since [this commit](https://github.com/projectcontour/contour-operator/commit/337e206e63f6c8042c634d2024cdaf12e7e68b71). This issue is for adding golint-ci as part of the make `test` target so it can be run locally. golint-ci has different installs for running in CI and running locally.

stevesloka:
We have this already inside: `make check` (https://github.com/projectcontour/contour-operator/blob/main/Makefile#L75)

I think I forgot to tag with the PR (#93).

PR #118, opened 2020-11-16
Updates Make Target in Contrib Doc

danehans:
Updates the contrib guide to use the `check` target.

xref: https://github.com/projectcontour/contour-operator/issues/9

/assign @stevesloka @skriss @jpeach 
/cc @Miciah

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -44,7 +44,7 @@ This produces a `contour-operator` binary in your `$GOPATH/bin` directory and ru
 You can run all the unit tests for the project:
 
 ```
-make test
+make check
 ```
 
 To run the tests for a single package, change to package directory and run:


skriss (approved):
LGTM

Makefile
```diff
@@ -64,12 +64,8 @@ endif
 
 all: manager
 
-# Run tests & validate against linters
-.PHONY: check
-check: test lint-golint
-
 # Run tests
-test: generate fmt vet manifests
+test: generate fmt vet lint-golint manifests
```
stevesloka, line 68: FYI, this change is going to run golint-ci twice now, once in `make test` as well as once for the GH action (https://github.com/projectcontour/contour-operator/blob/main/.github/workflows/prbuild.yaml#L25-L33).

danehans: @stevesloka thanks for catching this. Commit `7db0230` keeps the existing make targets and simply updates the contrib guide to use the appropriate target.

stevesloka: If you prefer to use `make test` we can swap things around. I'll merge this and can follow up with another PR if desired.

stevesloka (approved):
/lgtm

PR #127, opened 2020-11-18
Updates Makefile Uninstall Target

danehans:
Adds the Contour CRDs to the `uninstall` make target. This is needed since the `install` make target installs these CRDs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,11 +88,13 @@ run: generate fmt vet manifests
 install: manifests
 	kustomize build config/crd | kubectl apply -f -
 	# Remove the following when https://github.com/projectcontour/contour-operator/issues/42 merges.
-	kubectl apply -f https://raw.githubusercontent.com/projectcontour/contour/main/examples/contour/01-crds.yaml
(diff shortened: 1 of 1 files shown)


stevesloka (approved):
/lgtm

Makefile
```diff
@@ -93,6 +93,8 @@ install: manifests
 # Uninstall CRDs from a cluster
 uninstall: manifests
 	kustomize build config/crd | kubectl delete -f -
+	# Remove the following when https://github.com/projectcontour/contour-operator/issues/42 merges.
+	kubectl delete -f https://raw.githubusercontent.com/projectcontour/contour/main/examples/contour/01-crds.yaml
```
stevesloka, line 97: The creation of this (https://github.com/projectcontour/contour-operator/pull/120/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R114) has a `VERSION` variable in the url. Should this have that as well?

stevesloka (changes requested):

Makefile
danehans: @stevesloka good call. Commit `6c00514` resolves this comment.

stevesloka (approved):

PR #129, opened 2020-11-19
Fixes Image References in Operator Deployment

danehans:
This PR fixes a bug introduced by https://github.com/projectcontour/contour-operator/pull/116 that causes the operator use image `contour` instead of `contour-operator`.


danehans:
@stevesloka future instances of this issue should be prevented by https://github.com/projectcontour/contour-operator/pull/130.

skriss (approved):

PR #130, opened 2020-11-19
Adds the Deploy Target to CI

danehans:
Adds the `deploy` make target to the `test-linux` GitHub Actions job.

Requires: https://github.com/projectcontour/contour-operator/pull/120

### DIFF
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -21,7 +21,7 @@ jobs:
       - name: test
         run: |
           go mod vendor
-          make local-cluster install test
+          make local-cluster install deploy test
   codespell:
     name: Codespell
     runs-on: ubuntu-latest


danehans:
Note that this PR will prevent future instances of https://github.com/projectcontour/contour-operator/pull/129.

skriss:
@danehans CI is failing - see https://github.com/projectcontour/contour-operator/pull/130/checks?check_run_id=1426589913#step:5:130 - it looks like the namespace name is getting duplicated. I wonder if that's due to https://github.com/projectcontour/contour-operator/blob/main/config/default/kustomization.yaml#L9 which is getting applied to the `Namespace` name itself?

danehans:
@skriss thanks for looking into this issue. I did not observe this behavior locally b/c I was running a newer version of kustomize. I installed kustomize `v3.5.4` as specified in `install-kubernetes-toolchain.sh` and reproduced the issue. I submitted https://github.com/projectcontour/contour-operator/pull/132 as a fix to the issue, PTAL.

skriss (approved):
LGTM

PR #132, opened 2020-11-23
Removes contour-operator Config Prefix

danehans:
Previously, kustomize would add the `contour-operator` prefix to resources, including the `contour-operator` namespace. This is not an issue with newer versions of kustomize that do not use the `namePrefix` with namespaces, but is with `v3.5.4` used by `install-kubernetes-toolchain.sh`. This PR removes `namePrefix` and renames config resources based on the `contour-operator` naming scheme.

xref: https://github.com/projectcontour/contour-operator/pull/130

### DIFF
--- a/config/rbac/contour_viewer_role.yaml
+++ b/config/rbac/contour_viewer_role.yaml
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: contour-viewer-role
+  name: contour-viewer
 rules:
 - apiGroups:
   - operator.projectcontour.io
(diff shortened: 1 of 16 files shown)


config/rbac/contour_viewer_role.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: contour-viewer-role
+  name: contour-viewer
```
jpeach, line 5: Why not `contour-operator-viewer`?

danehans: @jpeach b/c the role is to view the `Contour` custom resource. The operator's service account will be given this cluster role so it can view `Contour` custom resources.

jpeach (approved):

PR #133, opened 2020-11-24
Adds e2e Testing

danehans:
Adds e2e tests that include:
- `.github/workflows/prbuild.yaml`: Adds the `test-e2e` make target to action `test`.
- `Makefile`: Adds `test-e2e` target to run e2e tests.
- `docs/CONTRIBUTING.md`: Updates testing references in contrib docs.
- `test/e2e/operator_test.go`: Adds e2e tests.
- `test/e2e/utils.go`: Adds e2e test utility functions.
- `util/object.go`: Adds functions for creating new objects used by e2e tests.

Requires: https://github.com/projectcontour/contour-operator/pull/132

### DIFF
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -21,7 +21,7 @@ jobs:
       - name: test
         run: |
           go mod vendor
-          make local-cluster install deploy test
+          make local-cluster test test-e2e
   codespell:
     name: Codespell
     runs-on: ubuntu-latest
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -40,9 +40,9 @@ make manager
 
 This produces a `contour-operator` binary in your `$GOPATH/bin` directory and runs go fmt and go vet against the code.
 
-### Running the unit tests
+### Running the tests
 
-You can run all the unit tests for the project:
+To run all the unit tests for the project:
 
 ```
 make check
@@ -54,7 +54,20 @@ To run the tests for a single package, change to package directory and run:
 go test .
 ```
 
-__Note:__ Unit tests must pass for your PR to get merged.
+The e2e tests require a Kubernetes cluster and uses your current cluster context.
+To create a [kind](https://kind.sigs.k8s.io/) Kubernetes cluster:
+
+```
+make local-cluster
+```
+
+To run the e2e tests for the project:
+
+```
+make test-e2e
+```
+
+__Note:__ Unit and e2e tests must pass for your PR to get merged.
 
 ## Contribution workflow
 
@@ -169,8 +182,7 @@ Operator CRDs and run the operator in the foreground:
 make run
 ```
 
-Before submitting your changes, follow the image-based deployment instructions to ensure the operator works as
-expected within a Kubernetes cluster.
+Before submitting your changes, run the [required tests](#Running-the-tests).
 
 ## DCO Sign off
 
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -1,0 +1,123 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build e2e
+
+package e2e
+
+import (
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	operatorv1alpha1 "github.com/projectcontour/contour-operator/api/v1alpha1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	// kclient is the Kubernetes client used for e2e tests.
+	kclient client.Client
+	// ctx is an empty context used for client calls.
+	ctx = context.TODO()
+	// operatorName is the name of the operator.
+	operatorName = "contour-operator"
+	// operatorNs is the name of the operator's namespace.
+	operatorNs = "contour-operator"
+	// defaultContourNs is the default spec.namespace.name of a Contour.
+	defaultContourNs = "projectcontour"
+	// testUrl is the url used to test e2e functionality.
+	testUrl = "http://local.projectcontour.io/"
+	// expectedDeploymentConditions are the expected status conditions of a
+	// deployment.
+	expectedDeploymentConditions = []appsv1.DeploymentCondition{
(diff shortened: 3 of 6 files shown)


docs/CONTRIBUTING.md
```diff
@@ -54,7 +54,13 @@ To run the tests for a single package, change to package directory and run:
...
 ```
 
-__Note:__ Unit tests must pass for your PR to get merged.
+To run the e2e test for the project:
+
+```
+make test-e2e
+```
+
+__Note:__ Unit and e2e tests must pass for your PR to get merged.
```
jpeach, line 63: Should mention that this requires an existing Kubernetes cluster and uses your current cluster context.

test/e2e/utils.go
```diff
@@ -0,0 +1,252 @@
...
+		_, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return false, nil
+		}
+		return true, nil
+	})
+	if err != nil {
+		t.Fatalf("failed to curl %s: %v", url, err)
+	}
+	t.Logf("curl'd url %s", url)
```
jpeach, line 214: These errors are misleading since `curl` isn't actually involved. Suggest something similar to:

```Go
t.Logf("%s %q returned status %d", method, url, status)
```

same for the `Fatalf` above

test/e2e/utils.go
```diff
@@ -0,0 +1,252 @@
...
+		Scheme: scheme,
+	}
+	kubeClient, err := client.New(config.GetConfigOrDie(), opts)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create kube client: %v\n", err)
+	}
+	return kubeClient, nil
+}
+
+func newContour(t *testing.T, ctx context.Context, cl client.Client, name, ns string) {
```
jpeach, line 48: Everything in this file that takes a `testing.T` should first call `t.Helper()`. This makes sure that the test failure is attributed to the caller, not the the helper function, which makes it much easier to triage test failures.

docs/CONTRIBUTING.md
jpeach: Mention that you need to build the kind cluster with `./hack/kind-dev-cluster.sh`

test/e2e/operator_test.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+func TestDefaultContour(t *testing.T) {
+	testName := "test-default-contour"
+	newContour(t, ctx, kclient, testName, operatorNs)
+	defer deleteContour(t, ctx, kclient, 3*time.Minute, testName, operatorNs)
+	// Defer the namespace deletion since contour.spec.namespace.removeOnDeletion
+	// defaults to false.
+	defer deleteNamespace(t, ctx, kclient, 3*time.Minute, defaultContourNs)
```
jpeach, line 62: when I tried this this namespace deletion fails because (I think) the default operator deployment was still present

```
=== RUN   TestDefaultContour
    utils.go:214: curl'd url http://local.projectcontour.io/
    utils.go:114: timed out waiting for namespace projectcontour to be deleted: timed out waiting for the condition
    utils.go:81: deleted contour contour-operator/test-default-contour
--- FAIL: TestDefaultContour (233.10s)
...
$ k get deployments -A
NAMESPACE            NAME                     READY   UP-TO-DATE   AVAILABLE   AGE
contour-operator     contour-operator         1/1     1            1           7m22s
```

test/e2e/operator_test.go
```diff
@@ -0,0 +1,82 @@
...
+	t.Logf("observed available status condition for contour %s/%s", testName, operatorNs)
+
+	appName := "kuard"
+	newDeployment(t, ctx, kclient, appName, defaultContourNs, "gcr.io/kuar-demo/kuard-amd64:1", 3)
+	if err := waitForDeploymentStatusConditions(t, ctx, kclient, 3*time.Minute, appName, defaultContourNs, expectedDeploymentConditions...); err != nil {
+		t.Errorf("failed to observe expected conditions for deployment %s/%s: %v", defaultContourNs, appName, err)
+	}
+	newClusterIPService(t, ctx, kclient, appName, defaultContourNs, 80, 8080)
+	newIngress(t, ctx, kclient, appName, defaultContourNs, appName, 80)
+	waitForCurl(t, testUrl, 1*time.Minute)
```
jpeach, line 81: In general, I think it's a better approach to write utilities to return errors, then simply use `assert.NoError` in the test caller. That ends up generating better error context in most cases. Not a blocker for this review, but maybe plan to revisit in the future.

jpeach (changes requested):
Let's address the docs and make sure the e2e test passes out of the box. Everything else can be deferred if you prefer.

test/e2e/operator_test.go
danehans: The operator is running in ns `contour-operator` and the test is deleting the operand ns, i.e. `projectcontour`. Did `projectcontour` eventually get deleted? At times I've seen namespace deletion take a while. The instance of Contour, `contour-operator/test-default-contour` should be deleted first and then the namespace should be deleted. Otherwise the operator may be trying to recreate the `projectcontour` namespace. I may remove `defer deleteNamespace()` since it's currently unneeded. It can be reintroduced as additional e2e tests are added.

.github/workflows/prbuild.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: test
         run: |
           go mod vendor
-          make local-cluster install deploy test
+          make local-cluster test test-e2e
```
skriss, line 24: in the future you might consider splitting these into separate jobs so they can run in parallel, but it's probably not a big deal right now.

test/e2e/operator_test.go
```diff
@@ -0,0 +1,100 @@
+// +build e2e
```
skriss, line 1: need to add copyright header (FYI, we recently added a lint check for this in Contour, would be good to do the same here: https://github.com/projectcontour/contour/pull/3133)

test/e2e/utils.go
```diff
@@ -0,0 +1,196 @@
+// +build e2e
```
skriss, line 1: need to add copyright header

skriss (approved):
I tried this locally and had everything pass once I fixed my local controller-gen version. Couple minor comments.

danehans:
@jpeach commit `8ca8067` resolves your comments above, PTAL.

test/e2e/utils.go
danehans: @skriss commit `644695b` resolves this comment.

test/e2e/operator_test.go
danehans: @skriss commit `644695b` resolves this comment. I created https://github.com/projectcontour/contour-operator/issues/140 to address the boilerplate check.

.github/workflows/prbuild.yaml
danehans: @skriss that's a great idea. I'd like to keep `prbuild.yaml` as-is for this PR and I'll create a follow-on PR to split the jobs.

config/manager/kustomization.yaml
```diff
@@ -5,4 +5,4 @@ resources:
 images:
 - name: controller
   newName: docker.io/projectcontour/contour-operator
-  newTag: main
+  newTag: bbfcec2c
```
skriss, line 8: guessing this should be reverted :)

danehans: @skriss thanks for catching, I reverted in commit `a154c90`.

skriss (approved):

jpeach (approved):

Issue #13, opened 2020-09-02
Evaluate moving unit tests to stretchr/testify


danehans:
**Please describe the problem you have**
Evaluate whether standardizing on the stretchr/testify framework for unit tests is worth adding as an external dependency.

/cc @jpeach

youngnick:
We use it in mainline Contour, so I'm supportive.

jpeach:
> We use it in mainline Contour, so I'm supportive.

Yeh we don't use the whole framework, but the `require` and `assert` packages are pretty handy and save a bunch of tedious test code.

PR #147, opened 2020-12-02
Adds Spec Namespace e2e Test

danehans:
Adds an e2e test to exercise `spec.namespace.name` and `spec.namespace.removeOnDeletion` fields of a `Contour`.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -17,6 +17,7 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


test/e2e/operator_test.go
```diff
@@ -121,3 +132,57 @@ func TestDefaultContour(t *testing.T) {
 		t.Fatalf("failed to delete namespace %s: %v", defaultContourNs, err)
 	}
 }
+
+func TestContourSpecNs(t *testing.T) {
+	testName := "test-user-contour"
+	specNs := fmt.Sprintf("%s-%s", defaultContourNs, testName)
+	removeNs := true
+	cntr, err := newContour(ctx, kclient, testName, operatorNs, specNs, removeNs)
+	if err != nil {
```
skriss, line 141: might be worth pulling in the [testify](https://github.com/stretchr/testify) `assert` and `require` packages here -- helps cut down the error-checking verbosity a bit. `assert` will log errors, and `require` will use fatal (log and exit immediately). Then all the `if err != nil { ...` checks can be replaced with `require.NoError(t, err, <msg>)`.

skriss (approved):
LGTM, one suggestion that can be tackled as a followup.

test/e2e/operator_test.go
danehans: Yeah, we have an issue created for using testify https://github.com/projectcontour/contour-operator/issues/13.

Issue #41, opened 2020-10-11
Validate Image Flags


danehans:
**Please describe the problem you have**
The value of the `--image` flag is currently accepted as an opaque string. The image value should be validated according to the Docker [reference type](https://pkg.go.dev/github.com/docker/distribution/reference). The same should be accomplished for all image references.

/cc @jpeach

PR #149, opened 2020-12-02
Validates Image Command-line Flags

danehans:
Validates `contour-image` and `envoy-image` command-line flags.

Fixes: https://github.com/projectcontour/contour-operator/issues/41

### DIFF
--- a/cmd/contour-operator.go
+++ b/cmd/contour-operator.go
@@ -20,6 +20,7 @@ import (
 	operatorv1alpha1 "github.com/projectcontour/contour-operator/api/v1alpha1"
 	contourcontroller "github.com/projectcontour/contour-operator/controller/contour"
 	"github.com/projectcontour/contour-operator/controller/manager"
+	oputil "github.com/projectcontour/contour-operator/util"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -30,8 +31,7 @@ import (
 )
 
 var (
-	scheme   = runtime.NewScheme()
-	setupLog = ctrl.Log.WithName("setup")
+	scheme = runtime.NewScheme()
 )
 
 func init() {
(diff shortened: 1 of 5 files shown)


Miciah:
You need to import hash algorithms to support digests:
```go
import (
	// ...

	_ "crypto/sha256"
	_ "crypto/sha512"

	// ...
)
```
I noticed this issue when I tried using a SHA512 digest (`<image>@sha256:<hash>`).  See <https://github.com/opencontainers/go-digest#usage>.  (I wonder if `github.com/docker/distribution/reference` has any other little surprises.)

danehans:
Commit `10db8c9` resolves https://github.com/projectcontour/contour-operator/pull/149#issuecomment-740259360.

skriss (approved):
LGTM

cmd/contour-operator.go
```diff
@@ -14,13 +14,16 @@
 package main
 
 import (
+	_ "crypto/sha256"
+	_ "crypto/sha512"
```
jpeach, line 18: Please comment why this is needed

cmd/contour-operator.go
```diff
@@ -61,6 +64,15 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
+	images := []string{contourImage, envoyImage}
+	for _, image := range images {
+		_, err := reference.Parse(image)
+		if err != nil {
+			setupLog.Error(err, "invalid image reference", "value", image)
```
jpeach, line 71: While we are touching this, let's make `setupLog` local, since it doesn't need to be a global var.

cmd/contour-operator.go
```diff
@@ -61,6 +64,15 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
+	images := []string{contourImage, envoyImage}
+	for _, image := range images {
+		_, err := reference.Parse(image)
```
jpeach, line 69: Consider wrapping `reference.Parse` and adding a test to defend against the fragility that @Miciah pointed out.

jpeach (changes requested):

danehans:
@jpeach commit `8ccfb9e` resolves your feedback above, PTAL.

stevesloka (approved):
/lgtm

PR #150, opened 2020-12-03
DO NOT MERGE: Updates Services for Port Names

danehans:
Updates the Contour and Envoy service to use named target ports. Names provide greater flexibility for supporting different container networking use cases According to k8s [docs](https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service):
> Port definitions in Pods have names, and you can reference these names in the targetPort attribute of a Service. This works even if there is a mixture of Pods in the Service using a single configured name, with the same network protocol available via different port numbers. This offers a lot of flexibility for deploying and evolving your Services. For example, you can change the port numbers that Pods expose in the next version of your backend software, without breaking clients.

### DIFF
--- a/controller/contour/daemonset.go
+++ b/controller/contour/daemonset.go
@@ -63,6 +63,10 @@ const (
 	envoyDaemonSetLabel = "contour.operator.projectcontour.io/daemonset-envoy"
 	// xdsResourceVersion is the version of the Envoy xdS resource types.
 	xdsResourceVersion = "v3"
+	// httpPortName is the name of an http container port.
+	httpPortName = "http"
+	// httpsPortName is the name of an https container port.
+	httpsPortName = "https"
 )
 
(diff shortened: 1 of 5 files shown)


stevesloka (approved):
I've never been a fan of port names, to me the actual port numbers are more straight forward to understand what is actually in use. Just my 2 cents.

danehans:
> I've never been a fan of port names, to me the actual port numbers are more straight forward to understand what is actually in use. Just my 2 cents.

@stevesloka My thinking is that using port name references will make easier to implement https://github.com/projectcontour/contour-operator/pull/151. I'll put a hold on this PR and reopen if this ends up being the case.

danehans:
https://github.com/projectcontour/contour-operator/pull/152 does not use port names, so I'm closing this PR.

Issue #70, opened 2020-10-22
Configurable Contour and Envoy Network Endpoint Publishing


danehans:
**Please describe the problem you have**
Make Contour and Envoy network endpoints configurable. For example, I may not want to use a LB service to expose Envoy (default). I may want to only use a ClusterIP or a range of other possibilities for Contour and Envoy.

/cc @jpeach @Miciah

danehans:
xref: https://kubernetes.slack.com/archives/C8XRH2R4J/p1603393745340600

Miciah:
Right now, the examples on projectcontour/contour use host ports.  Host ports are not well documented.  Using host networking (which is better documented) solves the same problem for bare metal, and using node ports is preferable on cloud platforms (and on bare metal when an external load-balancer is used).  

I believe we should change the projectcontour/contour examples (a) to use host networking instead of host ports to satisfy the bare-metal use-case and (b) define a LoadBalancer-type service (which doesn't necessarily need host ports or host networking, but should be compatible with either) to satisfy the cloud use-case.  We should propose this as a change to the projectcontour/contour examples first, and then update projectcontour/contour-operator once we have made the change in projectcontour/contour.  

We'd need to verify that changing host ports to host networking does not break kind.  We'd also need to sort out any implications for SCCs (there may not be an issue for vanilla Kubernetes).  

As a follow-up, we can change contour-operator (a) to turn host networking on and omit the LB service on bare-metal platforms and (b) to turn host networking off and create the LB service on cloud platforms.

danehans:
xref https://github.com/projectcontour/contour-operator/issues/105 to make ports configurable. containerPorts should be configurable. `nodePort` of the envoy service should be configurable when using a NodePort service.

PR #151, opened 2020-12-04
Adds Network Publishing API

danehans:
Updates the `Contour` API type to support network endpoint management.

__Note:__ The API design supports managing Envoy network endpoints with the ability to add Contour network endpoints in the future.

Fixes: https://github.com/projectcontour/contour-operator/issues/70


api/v1alpha1/contour_types.go
```diff
@@ -89,6 +96,49 @@ type NamespaceSpec struct {
 	RemoveOnDeletion bool `json:"removeOnDeletion,omitempty"`
 }
 
+// EndpointPublishing defines the schema to publish network endpoints and represents
+// the publishing type.
+type EndpointPublishing struct {
```
danehans, line 101: In the future, fields will be added to support config knobs for the different endpoint publishing types.  `EndpointPublishing` will become a union type and `Type` will be the union discriminator.

api/v1alpha1/contour_types.go
```diff
@@ -66,6 +66,13 @@ type ContourSpec struct {
 	//
 	// +kubebuilder:default={name: "projectcontour", removeOnDeletion: false}
 	Namespace NamespaceSpec `json:"namespace,omitempty"`
+
+	// EndpointPublishing is used to publish the network endpoints of Envoy.
+	//
+	// If unset, defaults to LoadBalancerService.
+	//
+	// +kubebuilder:default={type: "LoadBalancerService"}
+	EndpointPublishing EndpointPublishing `json:"endpointPublishing,omitempty"`
```
jpeach, line 75: I don't love this name, and I wonder whether we can come up with something more evocative. Data plane strategy? Envoy Service type?

jpeach (approved):
Approving, since this seems like a very reasonable API to me. Not merging yet, since I think the `EndpointPublishing` nomenclature is confusing and I'd like to give people a chance to suggest alternatives.

api/v1alpha1/contour_types.go
danehans: @jpeach I agree, I'm not crazy about the name, but I'm struggling with other names. This API is about publishing the network endpoints of Envoy.

danehans:
@jpeach during yesterday's community meeting we discussed whether it's necessary to manage the network endpoints of Contour. @stevesloka provided a use-case where Contour runs in a separate cluster from the managed Envoys, but we agreed that's not applicable since each cluster would run contour-operator. If Contour network endpoint publishing is not required today, should this API be designed to support the use case in the future. For example:

__Current__
```
spec:
  endpointPublishing:
    type: LoadBalancerService # Publishes Envoy's network endpoints using a k8s LB svc.
```

Here is an example of an API that can support Envoy (`dataPlane`) and Contour (`controlPlane`) network endpoint publishing.
__Option 1__
```
spec:
  endpointPublishing:
    dataPlane:
      type: LoadBalancerService # Publishes Envoy's network endpoints using a k8s LB svc.
    # future (if needed)
    controlPlane:
      type: LoadBalancerService # Publishes Contour's network endpoints using a k8s LB svc.
```

Here is an example of an API that abstracts the control and data planes at `spec`:
__Option 2__
```
spec:
  controlPlane:
    endpointPublishing:
      type: LoadBalancerService # Publishes Envoy's network endpoints using a k8s LB svc.
  # future (if needed)
  dataPlane:
    endpointPublishing:
      type: LoadBalancerService # Publishes Envoy's network endpoints using a k8s LB svc.
```

I anticipate future data-plane and control-plane specific requirements that the API will need to support. I'm beginning to think it makes sense to separate data/control planes at spec. All other spec-level fields will be applicable to both the control and data planes. Thoughts? cc: @stevesloka @skriss @Miciah

danehans:
I created [this](https://docs.google.com/spreadsheets/d/12vwHjRH0JgpRNl82QXt_GtUgxhAgRwa4khmFOSCYpbo/edit?usp=sharing) sheet to get input from the community for naming the API.

api/v1alpha1/contour_types.go
```diff
@@ -89,6 +96,49 @@ type NamespaceSpec struct {
 	RemoveOnDeletion bool `json:"removeOnDeletion,omitempty"`
 }
 
+// EndpointPublishing defines the schema to publish network endpoints and represents
+// the publishing type.
+type EndpointPublishing struct {
+	// Type is the type of publishing strategy to use. Valid values are:
+	//
+	// * LoadBalancerService
```
stevesloka, line 104: nit: Should we drop `Service` from these names and just call them `LoadBalancer`, or `NodePort` to match Kubernetes?

api/v1alpha1/contour_types.go
```diff
@@ -89,6 +96,49 @@ type NamespaceSpec struct {
...
+// +kubebuilder:validation:Enum=LoadBalancerService;NodePortService
+type EndpointPublishingType string
+
+const (
+	// LoadBalancerService publishes an endpoint using a Kubernetes LoadBalancer
+	// Service.
+	LoadBalancerServiceStrategyType EndpointPublishingType = "LoadBalancerService"
+
+	// NodePortService publishes an endpoint using a Kubernetes NodePort Service.
+	NodePortServiceStrategyType EndpointPublishingType = "NodePortService"
```
stevesloka, line 139: Maybe add `ClusterIP` for those who might need an internal ingress controller not exposed outside the cluster?

danehans: @stevesloka I agree that `ClusterIP` or `ClusterIPService` will be a supported type. I would like to keep the initial API represented in this PR as minimal as possible and add types as needed. To provide parity with Contour's rendered manifest, this initial API needs to only support `LoadBalancerService` and `NodePortService`.

danehans:
@jpeach @stevesloka although a k8s `Service` will back the `EndpointPublishing` types in the initial API, other non `Service`-based types may be added in the future. For example, `HostNetwork` would set `hostNetwork: true` for the Envoy `DaemonSet` and not create a Service. I agree that `Endpoint` may cause confusion, so let me see try to figure out a better name. Feel free to provide any naming suggestions in https://github.com/projectcontour/contour-operator/pull/151#issuecomment-738935719.

api/v1alpha1/contour_types.go
danehans: @stevesloka as I mention in https://github.com/projectcontour/contour-operator/pull/151#issuecomment-742161550, future types may not use a k8s `Service` resource. I thought it made sense to use types that are backed by a k8s service to include "Service" in the name. Take `LoadBalancerService` for example, what if a future `LoadBalancer` type is added that creates a hardware load balancer outside the k8s cluster?

danehans:
@stevesloka @jpeach @Miciah commit `0652869` renames the API to avoid confusion with the k8s `Endpoints` API. The commit also redesigns the API to support managing Contour network publishing (if the need arises). Here are a few examples of a Contour with the proposed API changes:

__A Default Contour__
```
kind: Contour
metadata:
  name: contour-sample
spec: {} # Envoy will use container ports 80/443 and be published using a k8s LB service.
```

__A Contour with a NodePort Service__
```
kind: Contour
metadata:
  name: contour-sample
spec:
  networkPublishing:
    envoy:
      type: NodePortService
```

__A Contour with Non-Standard Container Ports__
```
kind: Contour
metadata:
  name: contour-sample
spec:
  networkPublishing:
    envoy:
      httpContainerPort: 8080
      httpsContainerPort: 8443
```

api/v1alpha1/contour_types.go
```diff
@@ -91,6 +101,81 @@ type NamespaceSpec struct {
 	RemoveOnDeletion bool `json:"removeOnDeletion,omitempty"`
 }
 
+// NetworkPublishing defines the schema for publishing Contour to a network.
+type NetworkPublishing struct {
```
youngnick, line 105: Sorry to be the late drive-by guy, but doesn't this essentially create a Gateway ( in service APIs speak?) Should we reuse that naming? `GatewayPublishing` perhaps?

youngnick (approved):
I'm in a similar boat to @jpeach on this one. I think the API itself looks great, but suggest that we reuse the Service APIs naming and go with `GatewayPublishing` instead.

I can certainly be talked out of that, but I think we're heading that way anyway, so let's just call it the same thing.

api/v1alpha1/contour_types.go
danehans: An instance of `Contour` creates a `Gateway` in Service API terms. Since this API is not landing in v1.11.0, maybe it's better to implement `NetworkPublishing` from a Service APIs perspective. I added "WIP" to this PR to hold it while I create a Service APIs design doc for the operator.

youngnick: I think we should build towards where we want to go; even in the event that we're not accepting the Service APIs objects (as in v1.11.0), the thing can still be called a Gateway reasonably accurately right now (based on a naive reading of the term). So, by calling it `GatewayPublishing`, we make something that's no more confusing that `NetworkPublishing` right now, and save ourselves needing to change it later. That seems better to me. Happy to be convinced otherwise though.

danehans: @youngnick the current API is designed to support more than managing the network endpoints of Envoy. In the future, a field can be added to manage the network endpoints of Contour. A Gateway may have multiple network endpoints, so I think network publishing is still applicable in the future when Service APIs is supported. @jpeach @stevesloka @Miciah any thoughts here?

stevesloka: I think this works.

Quick questions, what if I wanted to add tolerations or nodeselectors to say the Envoy Daemonset? Where would I do that? 

Could we add more options to this? Maybe that moves it from a ` networkPublishing` type to a `resourceConfiguration` or something? (Names are hard 😢 )

danehans: > Quick questions, what if I wanted to add tolerations or nodeselectors to say the Envoy Daemonset? Where would I do that?

@stevesloka the scope of this API should be specific to network publishing, i.e. protocol, port, service type, etc. We should create a separate API for handling placement. For example:
```yaml
kind: Contour
metadata:
  name: example-contour
  namespace: contour-operator
spec:
  nodePlacement:
    nodeSelector:
      matchLabels:
        foo: bar
```

Would we need to control placement of Contour and Envoy? If this is a feature that should be supported by the operator, can you create an issue and xref this comment?

stevesloka: Makes sense 👍

stevesloka (approved):
I think this looks good now pending any other comments (I know there's been a bunch of questions back & forth). =)

danehans:
Per https://github.com/projectcontour/contour-operator/pull/171, `Contour` will be utilized by Service APIs for `gatewayclass.spec.parametersRef`. Since this PR will be used for provisioning Contour using the `Contour` CRD or Service APIs, I would like to proceed with the `NetworkPublishing` API naming scheme. @youngnick @jpeach PTAL and let me know if you have any questions or concerns.

api/v1alpha1/contour_types.go
danehans: Updated the struct name to `NetworkPublishing`.

danehans:
Per https://github.com/projectcontour/contour/pull/3263, commit `9226ee0` updates the NetworkPublishing API to support a list of container ports instead of specifying individual fields for http and https container ports. This commit also updates the default port numbers to match upstream, i.e. 8443 for secure and 8080 for insecure. cc: @youngnick @stevesloka

api/v1alpha1/contour_types.go
```diff
@@ -91,11 +102,160 @@ type NamespaceSpec struct {
...
+	NodePortServicePublishingType NetworkPublishingType = "NodePortService"
+)
+
+// LoadBalancerStrategy holds parameters for a load balancer.
+type LoadBalancerStrategy struct {
+	// Scope indicates the scope at which the load balancer is exposed.
+	// Possible values are "External" and "Internal".
+	//
+	// +kubebuilder:default=External
+	Scope LoadBalancerScope `json:"scope,omitempty"`
```
stevesloka, line 189: Is the operator going to take on specific implementations (e.g. https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer). 

If the goal isn't to take those on then possibly a new `ProviderSetting` (or something) are a better spot to allow users to define those specific exceptions. We might need both to cover anyone who might need specific annotations but aren't yet supported in the operator.

stevesloka (approved):
Just one small question and it looks great to me! Thanks for the work on this @danehans! 🎉 

I think we should just implement it at this point and see how things shake out while we're still in alpha/beta versions.

api/v1alpha1/contour_types.go
danehans: @stevesloka yes, the operator will support provider-specific LB settings through `ProviderParameters`, i.e. LB connection tuning. All the major cloud providers support internal/external LB scoping, so I think it's appropriate to define scope outside `ProviderParameters`.

api/v1alpha1/contour_types.go
```diff
@@ -91,6 +98,155 @@ type NamespaceSpec struct {
...
+	Type NetworkPublishingType `json:"type,omitempty"`
+
+	// loadBalancer holds parameters for the load balancer. Present only if type is
+	// LoadBalancerService.
+	//
+	// If unspecified, defaults to an external Classic AWS ELB.
+	//
+	LoadBalancer LoadBalancerStrategy `json:"loadBalancer,omitempty"`
+
+	// ContainerPorts is a list of container ports to expose from the Envoy container(s).
```
danehans, line 151: @youngnick @stevesloka @skriss do you agree that at least 2 `ContainerPorts` must be defined to support Envoy's insecure and secure services? I'm unsure if it's possible or if a use-case exists to run the insecure/secure services on the same container port. Do you agree that 6 container ports is an appropriate maximum?

youngnick: I think those are totally acceptable numbers for now. We can change this later.

danehans:
Commit `f1fa042` changes MaxItem=2 and godocs for the containerPort field based on feedback from the maintainers call.

skriss (approved):
nothing to add here, LGTM

PR #152, opened 2020-12-04
Implements Network Publishing API

danehans:
Implements the Network Publishing API defined in https://github.com/projectcontour/contour-operator/pull/151.

Requires:
- https://github.com/projectcontour/contour-operator/pull/151
- https://github.com/projectcontour/contour-operator/pull/162

### DIFF
--- a/internal/equality/equality.go
+++ b/internal/equality/equality.go
@@ -164,9 +164,9 @@ func ClusterIPServiceChanged(current, expected *corev1.Service) (*corev1.Service
 	return updated, true
 }
 
-// LoadBalancerServiceChanged checks if the spec of current and expected match and if not,
-// returns true and the expected Service resource. The healthCheckNodePort and a port's
-// nodePort are not compared since they are dynamically assigned.
+// LoadBalancerServiceChanged checks if current and expected match and if not, returns
+// true and the expected Service resource. The healthCheckNodePort and a port's nodePort
+// are not compared since they are dynamically assigned.
 func LoadBalancerServiceChanged(current, expected *corev1.Service) (*corev1.Service, bool) {
 	changed := false
 	updated := current.DeepCopy()
@@ -213,6 +213,62 @@ func LoadBalancerServiceChanged(current, expected *corev1.Service) (*corev1.Serv
 
 	if !apiequality.Semantic.DeepEqual(current.Spec.Type, expected.Spec.Type) {
 		updated.Spec.Type = expected.Spec.Type
+		changed = true
+	}
+
+	if !apiequality.Semantic.DeepEqual(current.Annotations, expected.Annotations) {
+		updated.Annotations = expected.Annotations
+		changed = true
+	}
+
+	if !changed {
+		return nil, false
+	}
+
+	return updated, true
+}
+
+// NodePortServiceChanged checks if current and expected match and if not, returns
+// true and the expected Service resource. The healthCheckNodePort is not compared
+// since it's dynamically assigned.
+func NodePortServiceChanged(current, expected *corev1.Service) (*corev1.Service, bool) {
+	changed := false
+	updated := current.DeepCopy()
+
+	if len(current.Spec.Ports) != len(expected.Spec.Ports) {
+		updated.Spec.Ports = expected.Spec.Ports
+		changed = true
+	}
+
+	for i, p := range current.Spec.Ports {
+		if !apiequality.Semantic.DeepEqual(p, expected.Spec.Ports[i]) {
+			updated.Spec.Ports = expected.Spec.Ports
+			changed = true
+		}
+	}
+
+	if !apiequality.Semantic.DeepEqual(current.Spec.Selector, expected.Spec.Selector) {
+		updated.Spec.Selector = expected.Spec.Selector
+		changed = true
+	}
+
+	if !apiequality.Semantic.DeepEqual(current.Spec.ExternalTrafficPolicy, expected.Spec.ExternalTrafficPolicy) {
+		updated.Spec.ExternalTrafficPolicy = expected.Spec.ExternalTrafficPolicy
+		changed = true
+	}
+
+	if !apiequality.Semantic.DeepEqual(current.Spec.SessionAffinity, expected.Spec.SessionAffinity) {
+		updated.Spec.SessionAffinity = expected.Spec.SessionAffinity
+		changed = true
+	}
+
+	if !apiequality.Semantic.DeepEqual(current.Spec.Type, expected.Spec.Type) {
+		updated.Spec.Type = expected.Spec.Type
+		changed = true
+	}
+
+	if !apiequality.Semantic.DeepEqual(current.Annotations, expected.Annotations) {
+		updated.Annotations = expected.Annotations
 		changed = true
 	}
 
--- a/internal/object/object.go
+++ b/internal/object/object.go
@@ -117,13 +117,35 @@ func NewNamespace(name string) *corev1.Namespace {
 	}
 }
 
-// NewContour makes a Contour object using the provided ns/name
-// for the object's namespace and name.
-func NewContour(name, ns string) *operatorv1alpha1.Contour {
+// NewContour makes a Contour object using the provided ns/name for the
+// object's namespace/name, pubType for the network publishing type of
+// Envoy, and Envoy container ports 8080/8443.
+func NewContour(name, ns, specNs string, remove bool, pubType operatorv1alpha1.NetworkPublishingType) *operatorv1alpha1.Contour {
 	return &operatorv1alpha1.Contour{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: ns,
 			Name:      name,
+		},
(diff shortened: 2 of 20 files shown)


internal/equality/equality.go
```diff
@@ -215,6 +215,11 @@ func LoadBalancerServiceChanged(current, expected *corev1.Service) (*corev1.Serv
 		changed = true
 	}
 
+	if !apiequality.Semantic.DeepEqual(current.Annotations, expected.Annotations) {
```
skriss, line 218: nit: might want to tweak the function godoc since this isn't part of `spec`.

internal/object/object.go
```diff
@@ -115,14 +115,36 @@ func NewNamespace(name string) *corev1.Namespace {
 	}
 }
 
-// NewContour makes a Contour object using the provided ns/name
-// for the object's namespace and name.
-func NewContour(name, ns string) *operatorv1alpha1.Contour {
+// NewContour makes a Contour object using the provided ns/name for the
+// object's namespace/name, pubType for the network publishing type of
+// Envoy, and Envoy container ports 8080/8443.
+func NewContour(name, ns, specNs string, remove bool, pubType operatorv1alpha1.NetworkPublishingType) *operatorv1alpha1.Contour {
```
skriss, line 121: don't have to do it now, but we probably want to move to an options struct before long as this function's parameter list gets longer and longer.

internal/operator/controller/contour/daemonset.go
```diff
@@ -60,6 +60,10 @@ const (
 	envoyCfgFileName = "envoy.json"
 	// xdsResourceVersion is the version of the Envoy xdS resource types.
 	xdsResourceVersion = "v3"
+	// envoyInsecureContainerPort is the network port number of Envoy's insecure service.
```
skriss, line 63:
```suggestion
	// envoyInsecureContainerPort is the network port number of Envoy's insecure listener.
```

internal/operator/controller/contour/daemonset.go
```diff
@@ -60,6 +60,10 @@ const (
...
+	envoyInsecureContainerPort = int32(8080)
+	// envoySecureContainerPort is the network port number of Envoy's secure service.
```
skriss, line 65:
```suggestion
	// envoySecureContainerPort is the network port number of Envoy's secure listener.
```

internal/operator/controller/contour/service.go
```diff
@@ -152,34 +208,65 @@ func DesiredContourService(contour *operatorv1alpha1.Contour) *corev1.Service {
 
 // DesiredEnvoyService generates the desired Envoy Service for the given contour.
 func DesiredEnvoyService(contour *operatorv1alpha1.Contour) *corev1.Service {
+	var ports []corev1.ServicePort
+	for _, port := range contour.Spec.NetworkPublishing.Envoy.ContainerPorts {
+		var p corev1.ServicePort
+		httpFound := false
+		httpsFound := false
```
skriss, line 215: these don't look like they're ever being set

skriss (changes requested):

internal/operator/config/config.go
```diff
@@ -13,12 +13,14 @@
 
 package config
 
-var (
+const (
 	DefaultContourImage           = "docker.io/projectcontour/contour:main"
 	DefaultEnvoyImage             = "docker.io/envoyproxy/envoy:v1.17.0"
 	DefaultMetricsAddr            = ":8080"
```
stevesloka, line 19: I know this isn't part of this change but `8080` isn't the default metrics port. This would then conflict with the default http port.

internal/operator/controller/contour/service.go
```diff
@@ -43,6 +43,62 @@ const (
 	// TODO [danehans]: Make proxy protocol configurable or automatically enabled. See
 	// https://github.com/projectcontour/contour-operator/issues/49 for details.
 	awsLbBackendProtoAnnotation = "service.beta.kubernetes.io/aws-load-balancer-backend-protocol"
+	// envoyServiceHTTPPort is the HTTP port number of the Envoy service.
+	envoyServiceHTTPPort = 80
+	// envoyServiceHTTPSPort is the HTTPS port number of the Envoy service.
+	envoyServiceHTTPSPort = 443
+	// envoyNodePortHTTPPort is the NodePort port number for Envoy's HTTP service. For NodePort
+	// details see: https://kubernetes.io/docs/concepts/services-networking/service/#nodeport
+	envoyNodePortHTTPPort = 30080
```
stevesloka, line 52: Why set these statically? There's a possibility of conflict with another nodePort type service.

internal/operator/controller/contour/service.go
```diff
@@ -43,6 +43,62 @@ const (
...
+	// envoyNodePortHTTPSPort is the NodePort port number for Envoy's HTTPS service. For NodePort
+	// details see: https://kubernetes.io/docs/concepts/services-networking/service/#nodeport
+	envoyNodePortHTTPSPort = 30443
+	// awsProviderType is the name of the Amazon Web Services provider.
+	awsProviderType = "AWS"
```
stevesloka, line 57: Not a problem for now, but we should look at defining these out into different files with an interface or something maybe. That would make adding new one's easy. 

But not needed for this specific PR.

stevesloka (changes requested):

internal/operator/controller/contour/service.go
danehans: @stevesloka I'm trying to avoid exposing nodeport configuration during the initial release of the Network Publishing API. The API changes were already quite large without exposing nodeport configuration. I would like to revisit the topic when users identify nodeport configuration as an issue or a desired capability.

internal/object/object.go
danehans: Agreed. As you mention, let me refactor into an options struct in a follow-on PR.

internal/operator/controller/contour/daemonset.go
danehans: Ack

internal/operator/controller/contour/daemonset.go
danehans: Ack

internal/operator/controller/contour/service.go
danehans: These vars were removed in a different PR that recently merged.

internal/operator/config/config.go
danehans: @stevesloka `DefaultMetricsAddr` is the default metrics address for the operator (not Contour's metrics address).

danehans:
@stevesloka @skriss I addressed your feedback in my above comments or in commit `6d1e09b`.

internal/operator/controller/contour/service.go
skriss: I think you just want to set `httpFound` and `httpsFound` to `true` within the `case` statements for each of them. I'm not quite sure what you mean about the other PR since they were added in this PR.

internal/operator/config/config.go
stevesloka: Ahh you're right, got those mixed up. 🙃

internal/operator/controller/contour/service.go
stevesloka: Sure that's fine, but is it better to not define NodePorts and let Kubernetes define which it should use?

danehans: @stevesloka thank you for the feedback.  I originally thought it would be beneficial to have a well-known nodeport, but I created https://github.com/projectcontour/contour-operator/issues/184 to surface this information in status.  I have removed the static nodeport assignment for the Envoy Service, but kept the port assignments for the kind cluster script.

internal/operator/controller/contour/service.go
danehans: @skriss I apologize, I got lost in the review so please disregard my comment above. I have updated the PR so the conditional loops through the ports, sets the correct bool, and applies the service port params.

internal/operator/controller/contour/service.go
danehans: @stevesloka after further investigation, we need the ability to assign the nodeport through the Network Publishing API or statically set a port. This is because the kind config requires port mapping definitions:
```
  extraPortMappings:
  - containerPort: 30080
    hostPort: 80
    listenAddress: "0.0.0.0"
  - containerPort: 30443
    hostPort: 443
    listenAddress: "0.0.0.0"
```

Without these ports being used by a `Contour`, the e2e tests fail. I'm going to revert the latest commit so static nodeports are used and expose the nodeport configuration in a subsequent PR.

internal/object/object.go
danehans: xref: https://github.com/projectcontour/contour-operator/issues/189

danehans:
xref: https://github.com/projectcontour/contour-operator/issues/159

skriss (approved):
nothing else from me, LGTM pending @stevesloka's comments

stevesloka (approved):

PR #153, opened 2020-12-04
Fixes Operator Image Dev/Release Workflow

danehans:
Previously, using the `IMAGE` variable would not cause the image to be set in the operator's deployment. This PR allows kustomize to update the operator's image in `config/manager/manager.yaml` through the make `deploy` target.

Fixes a regression introduced by https://github.com/projectcontour/contour-operator/pull/129 and https://github.com/projectcontour/contour-operator/pull/132.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ all: manager
 check: test lint-golint lint-codespell
 
 # Run tests
-test: generate fmt vet manifests
+test: generate fmt vet manifests verify-image-refs
 	go test ./... -coverprofile cover.out
 
 lint-golint:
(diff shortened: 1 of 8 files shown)


jpeach (approved):
This seems OK. My one concern is that the `kustomize edit` makes unwanted on-disk changes that seems likely to accidentally get checked in. However that problem was already present, so if @danehans is OK with it then I guess I am too :)

danehans:
@jpeach thanks for the review. Let me know if you suggest a different approach. This is the same approach that was included in the kubebuilder bootstrap of the project. This fixes a regression introduced by https://github.com/projectcontour/contour-operator/pull/129 and https://github.com/projectcontour/contour-operator/pull/132. I plan on submitting a follow-on PR to add `config/manager/kustomization.yaml` to .gitignore.

danehans:
@jpeach I've added make targets to verify/reset the image reference in the files that get updated by using a custom image and/or tag. I have also updated the `deploy` target to use a script that sets the image tag to "main` if using the "docker.io/projectcontour/contour-operator" image. PTAL when you have a moment.

jpeach (approved):

PR #161, opened 2020-12-12
Use Kubernetes networking v1

### LINKED ISSUE #139: Use Kubernetes Networking v1
**Please describe the problem you have**
Currently, contour operator is using "k8s.io/api/networking/v1beta1".  The v1 package should be used instead.

sheetalsingala:
Signed-off-by: sheetalsingala <ssingala@redhat.com>
Fixes #139

### DIFF
--- a/util/object.go
+++ b/util/object.go
@@ -18,7 +18,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	networkingv1beta1 "k8s.io/api/networking/v1beta1"
+	networkingv1 "k8s.io/api/networking/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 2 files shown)


util/object.go
```diff
@@ -180,17 +180,21 @@ func NewClusterIPService(ns, name string, port, targetPort int) *corev1.Service
...
 		},
-		Spec: networkingv1beta1.IngressSpec{
-			Backend: &networkingv1beta1.IngressBackend{
-				ServiceName: backendName,
-				ServicePort: intstr.IntOrString{IntVal: int32(backendPort)},
+		Spec: networkingv1.IngressSpec{
+			DefaultBackend: &networkingv1.IngressBackend{
+				Service: &networkingv1.IngressServiceBackend{
+					Name: backendName,
+					Port: networkingv1.ServiceBackendPort{
```
danehans, line 194: @sheetalsingala instead of changing the signature of `NewIngress`, can you set `backendPort` to an int32, i.e. `int32(backendPort)` in the function?

sheetalsingala: I can do that. Thank you for the feedback

youngnick (approved):

Issue #160, opened 2020-12-12
Fix test-examples Make Target


danehans:
"Sending http request to http://local.projectcontour.io" iterates forever when the example deployment fails from `make test-examples`. The curl test should fail after a specified number of attempts or a timeout is reached.

danehans:
This is a good first issue for anyone interested.

sheetalsingala:
Hi @danehans, I see that there is already [timeout](https://github.com/projectcontour/contour-operator/blob/main/hack/test-examples.sh#L24) for `waitForHttpResponse`.
Am I looking at the right thing here?

danehans:
@sheetalsingala PTAL at https://github.com/projectcontour/contour-operator/pull/162 that includes the fix (among other improvements).

PR #162, opened 2020-12-14
Fixes Issues with Make Workflow

danehans:
Previously, the workflow was not using the generated operator image when running tests. This PR updates the workflow as follows:

1. Adds the `load-image` target for loading the generated operator image to a kind cluster for testing purposes.
2. Allows the `deploy` and `example` targets to use a tagged image other than `main`.
3. Sets the operator's `imagePullPolicy` to "Always". __Note:__ The `load-image` target will change the policy to "IfNotPresent" to allow kind nodes to serve the generated image.
4. Fixes an issues with the `test-example` target to exit properly when the test fails.

Fixes: https://github.com/projectcontour/contour-operator/issues/160


youngnick (approved):
LGTM, nice work.

jpeach (approved):

PR #164, opened 2020-12-16
build: always use the pinned controller-gen

jpeach:
When the build looks for an existing installation of controller-gen,
it can find one with a different version. This leads to spurious diffs
when the different version of controller-gen generates different output.

Better to use `go run` to run the versino that is pinned in the Go
modules. This means that everyone gets the same result.


stevesloka (approved):
/lgtm

PR #166, opened 2020-12-17
Sets the Operator Image Pull Policy to IfNotPresent

danehans:
https://github.com/projectcontour/contour-operator/pull/162 set `imagePullPolicy: Always` for the operator. This PR updates the value to `IfNotPresent` (default) to be consistent with Contour [examples](https://github.com/projectcontour/contour/tree/main/examples/contour).

### DIFF
--- a/hack/load-image.sh
+++ b/hack/load-image.sh
@@ -45,8 +45,6 @@ fi
 # the kind cluster. Set the pull policy with kustomize when
 # https://github.com/kubernetes-sigs/kustomize/issues/1493 is fixed.
 for file in config/manager/manager.yaml examples/operator/operator.yaml ; do
-  # The version might be main or OLDVERS depending on whether we are
-  # tagging from the release branch or from main.
   echo "setting \"imagePullPolicy: IfNotPresent\" for $file"
   run::sed \
     "-es|imagePullPolicy: Always|imagePullPolicy: IfNotPresent|" \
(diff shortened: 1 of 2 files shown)


jpeach (approved):

config/manager/manager.yaml
```diff
@@ -25,7 +25,7 @@ spec:
         args:
         - --enable-leader-election
         image: contour-operator
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
stevesloka, line 28: Should this be left as `Always` in a dev build scenario where you'd push the same image tag over and over testing?

skriss: In contour we leave the "source" YAML files as `Always` (for dev), then update it to `IfNotPresent` for the rendered file, which is the one that gets served for the quickstart link. It'd be nice to do the same here, I think.

danehans:
@stevesloka @skriss commit `12dcde1` resolves https://github.com/projectcontour/contour-operator/pull/166#discussion_r545193187. The pull policy will continue to be set as `Always`. When a release is cut, i.e. `hack/release/make-release-tag.sh`, the policy will be updated to `IfNotPresent`. PTAL when you have a moment.

stevesloka (approved):
/lgtm

skriss (approved):
LGTM :+1:

PR #171, opened 2021-01-05
Adds Service APIs Design Doc

danehans:
Adds a design doc that details support for Service APIs.


design/service-apis.md
```diff
@@ -0,0 +1,249 @@
...
+unspecified, the operator will create an instance of Contour with default settings. Since `parametersRef` does not
+contain a namespace field, the referenced resource must exist in a known namespace. The operator's namespace will be
+used as the known namespace. If the referenced resource cannot be found, the operator will set the GatewayClass
+`InvalidParameters` status condition to true.
+
+### ContourConfig
+
+A new ContourConfig CR will be introduced to provide a structured object for representing the Contour configuration.
+This configuration-only resource will be consumed by a GatewayClass. ContourConfig is equivalent to spec of a Contour CR
+to provide consistency and supportability for managing Contour using the Contour CR or Service APIs. For example:
```
danehans, line 82: Should `ContourConfig` be spec of `Contour` for consistency between the APIs or should we maintain an entirely separate API just to hold the Contour configuration for Service APIs? Note that when I say "configuration" I mean the e2e config of a COntour environment and not https://github.com/projectcontour/contour/blob/main/examples/contour/01-contour-config.yaml.

cc: @jpeach @stevesloka

design/service-apis.md
```diff
@@ -0,0 +1,249 @@
...
+Contour. Service APIs provides extension points to allow for custom resources to be linked at various layers of the API.
+This provides more granular customization at the appropriate places within the API structure. These extension points
+will be leveraged to expose Contour-specific configuration.
+
+## Detailed Design
+
+### GatewayClass
+
+GatewayClass is a cluster-scoped resource that defines a set of Gateways sharing a common behavior. A GatewayClass
+with `controller: projectcontour.io/contour-operator` will be managed by Contour Operator. There MUST be at least one
```
stevesloka, line 43: nit: Should this just be `projectcontour.io/contour`? The Gateway is handled Contour not the operator.

stevesloka: I guess the important bit is that is starts with `projectcontour.io`, the name after will be specific to the instance that should be managed, but all other names (i.e. not projectcontour.io will be ignored).

design/service-apis.md
```diff
@@ -0,0 +1,249 @@
...
+    kind: ContourConfig
+```
+
+The `controller` field is a domain/path string that indicates the controller responsible for managing Gateways of this
+class. The `projectcontour.io/contour-operator` string will be used to specify Contour Operator as the controller
+responsible for handling the GatewayClass. This field is not mutable and cannot be empty. Versioning can be accomplished
+by encoding the operator's version into the path. For example:
+
+```
+projectcontour.io/contour-operator/v1   // Use version 1
```
stevesloka, line 66: What version does this relate to? The verison of Contour? The version of the operator?

design/service-apis.md
stevesloka: Could they merge and remove the current config map as configuration? I think long term that's probably a better approach.

design/service-apis.md
```diff
@@ -0,0 +1,249 @@
...
+  labels:
+    app: example
+spec:
+  rules:
+  - forwardTo:
+    - serviceName: example-svc
+      port: 8080
+```
+
+The `addresses` field specifies one or more network addresses requested for this Gateway. If unspecified, an IP address
```
stevesloka, line 201: How does the operator determine if the address is invalid or unavailable?

design/service-apis.md
```diff
@@ -0,0 +1,249 @@
...
+
+Although Gateway supports multiple address types, only "IPAddress" will be supported by the Contour Operator.
+
+## Implementation Details
+
+### Controller
+
+Create a controller that reconciles the GatewayClass resource and performs the following:
+
+- Only reconcile a GatewayClass that specifies `controller: projectcontour.io/contour-operator`.
```
stevesloka, line 226: The operator should look for domains with `projectcontour.io/`, without needing the `contour-operator` bits I think. Is there a good reason to always add that? 

My thought is you could have one GatewayClass be `projectcontour.io/external` and another be `projectcontour.io/internal`.

stevesloka:
Also, how do we handle bits that aren't implemented? Like the example of the addresses field that only supports `IP`, just ignore or can we let the user know somewhere?

design/service-apis.md
danehans: > Could they merge and remove the current config map as configuration? 

@stevesloka even without adding support for Service APIs, the intent is to expose `contour.yaml` (i.e. Contour ConfigMap) configuration to users through the `Contour` custom resource. We want to be selective on what configuration we expose. This is why no `contour.yaml` was exposed through the API for v1.11.0.

design/service-apis.md
danehans: > Should this just be projectcontour.io/contour? The Gateway is handled Contour not the operator.

This field represents the controller that manages the `Gateway`. This controller will reside in Contour Operator so IMO the value should indicate `contour-operator`. According to https://github.com/projectcontour/contour/pull/3213/files and https://github.com/projectcontour/contour/issues/2809, Contour will manage `xRoute` resources and not `Gateway` resources.

design/service-apis.md
danehans: The field can be used to specify an operator version to use for gateways of this `GatewayClass`. Since the operator version is synchronized with the contour version, this field would also represent the Contour version.

design/service-apis.md
danehans: @stevesloka the operator will create the Envoy service with the provided address. The Service field that is set will depend on the network publishing type specified in the `ContourConfig`. For example, if the network publishing type is "Internal", the operator will create an Envoy service of type `ClusterIP` and set `clusterIPs: 1.2.3.4`. If service creation fails with "error: address unavailable" or "error: address in use", then the operator will surface a `Gateway` status condition.

design/service-apis.md
danehans: > The operator should look for domains with projectcontour.io/, without needing the contour-operator bits I think. Is there a good reason to always add that?

I believe so. See https://github.com/projectcontour/contour-operator/pull/171#discussion_r552937717.

> My thought is you could have one GatewayClass be projectcontour.io/external and another be projectcontour.io/internal.

The following workflow should support this use-case:
- Create a `GatewayClass` that references a `ContourConfig` containing the external config bits. The default settings should suffice since a Service of type "LoadBalancer" is used for Envoy.
- Create a `GatewayClass` that references a `ContourConfig` containing the internal config bits, i.e network publishing type "Internal" which creates an Envoy Service of type "ClusterIP" instead of type "LoadBalancer".
- Create a `Gateway` that references the external `GatewayClass`.
- Create a `Gateway` that references the internal `GatewayClass`.

You will now have two separate instances of Contour, each managing an Envoy DaemonSet.

danehans:
> Also, how do we handle bits that aren't implemented? Like the example of the addresses field that only supports IP, just ignore or can we let the user know somewhere?

@stevesloka we will need to document any deviations from Service APIs. Our intent should be to minimize the deviations over time.

As for the example of not supporting named addresses, may be we can with [externalName](https://kubernetes.io/docs/concepts/services-networking/service/#externalname). Are you aware of anyone using externalName for the Envoy Service?

design/service-apis.md
stevesloka: So today you'd have this: `projectcontour.io/contour-operator/v1.11.0`? Then you'd upgrade the operator to say v1.12.0 and you'd need to then change the gateway to be `projectcontour.io/contour-operator/v1.12.0` to pick up the update?

What happens with the older versions, would they be orphaned then since the operator only works on `latest`?

design/service-apis.md
stevesloka: ahh ok so you're going to surface the k8s service creation error back, that can work. I thought you were going to need to understand somehow if the address requested would work for each specific environment which seemed like a large scope.

design/service-apis.md
stevesloka: I guess my initial thought was having the name `contour-operator` in there. You explained that it's because the operator should manage which makes sense. What should we use then for say the quickstart for this field?

stevesloka:
@danehans no I'm not aware of anyone using externalName. My comment was more around surfacing back to the user they can't use a field, or something isn't yet implemented so they don't spent time troubleshooting something that just isn't yet implemented. Maybe set a status accordingly?

design/service-apis.md
stevesloka: It's more of a `Gateway` config vs a `ContourConfig`, maybe that name is what's confusing since now we have a contour configmap and a contour config CRD.

danehans: @stevesloka `gatewayclass.spec.parametersRef` is a reference to a resource containing implementation-specific config parameters. I chose `ContourConfig` instead of a name such as `GatewayConfig` since the parameters are implementation-specific.

design/service-apis.md
jpeach: I concur with @stevesloka. I think the class name should be `projectcontour.io/contour` to request that Contour satisfy the GatewayClass request. The fact that the operator provisions the class isn't relevant to the API.

Another way to say the same thing is that GatewayClass specifies the underlying infrastructure that gateways are scheduled onto. In this case, that infrastructure is Contour.

design/service-apis.md
```diff
@@ -0,0 +1,249 @@
...
+kind: GatewayClass
+apiVersion: networking.x-k8s.io/v1alpha1
+metadata:
+  name: example-gatewayclass
+spec:
+  controller: projectcontour.io/contour-operator
+  parametersRef:
+    name: example-contour-config
+    group: operator.projectcontour.io
+    kind: ContourConfig
```
jpeach, line 57: Can we re-use the `Contour` CRD here so that there's a common vocabulary?

danehans: @jpeach here are a few other approaches that we can consider:

1. Use a ConfigMap to represent the Contour-specific configuration. This ConfigMap would be referenced by `gatewayclass.spec.parametersRef`.
2. Add a field to `contour.spec`, that if set, will cause the operator to NOT instantiate a Contour environment. For example:
```
kind: Contour
metadata:
  name: contour-sample
spec:
  controller: GatewayClass
```

`controller: GatewayClass` indicates the `Contour` is managed by a gatewayclass controller and will cause the operator to NOT instantiate a Contour environment, while `controller: Operator` (default) will cause the operator to to  instantiate a Contour environment.

3. Similar to option 2, but use an annotation instead of adding a field to `Contour`.

The more I think about it, the more I like option 2. Unless I hear otherwise, I'll update the design based on option 2.

design/service-apis.md
danehans: @jpeach my understanding of the [spec](https://github.com/kubernetes-sigs/service-apis/blob/master/apis/v1alpha1/gatewayclass_types.go#L51-L52) is- who is responsible for managing Gateways of this class? The operator will be responsible for managing Gateways. Here's a sample use case:

1. A user creates GatewayClass/Gateway "Foo". Contour Operator creates the Contour environment and everything is great for a long time.
2. One day a new project called Foo comes along that is great at managing Contour using Service APIs.
3. The user is interested in creating a new Contour environment using project Foo. Since `projectcontour.io/contour` is used by Contour Operator and project Foo, they fight over which one is responsible for instantiating a Contour environment.

How can users differentiate using Contour Operator or another controller to manage Contour if both use the same `projectcontour.io/contour` value?

design/service-apis.md
danehans: > Then you'd upgrade the operator to say v1.12.0

@stevesloka this is the workflow that I envision:
- The infrastructure provider upgrades the operator image to v1.12.0, triggering an upgrade of the operator. The Contour CRD is updated if API changes are included in v1.12.0.
- The status of `projectcontour.io/contour-operator/v1.11.0` GatewayClasses are updated to "Waiting" and associated Gateways are updated to "NoSuchGatewayClass". The details for accomplishing this step is TBD and I welcome any input. Should Contour be responsible for updating GatewayClass/Gateway/xRoute status when it's in this "detached" state?
- The operator will no longer manage Gateways of `projectcontour.io/contour-operator/v1.11.0` GatewayClasses. These Gateways will continue to function and traffic will continue to be proxied. Contour will continue running v1.11.0 and Envoy v1.16.2.
- Optionally, the infrastructure provider updates the Contour resource referenced by the GatewayClass, or creates and associates a new Contour to the GatewayClass.
- The infrastructure provider updates Gateways by setting GatewayClasses to `projectcontour.io/contour-operator/v1.12.0`.  This triggers the operator to update Contour to v1.12.0 and optionally update the configuration. GatewayClass/Gateway/xRoute status conditions are updated accordingly.

__Note:__ GatewayClasses of `projectcontour.io/contour-operator` are equivalent to running the latest operator release. In this case, Gateways will be automatically upgraded when the operator is upgraded. `projectcontour.io/contour-operator` will be the only supported value for the initial support of Service APIs.

Thoughts?

design/service-apis.md
danehans: > What should we use then for say the quickstart for this field?

@stevesloka We can either version this field upon cutting a release, similar to how we manage the image versions, or we can omit the version from the GatewayClass so Contour Operator will always manage `projectcontour.io/contour-operator` GatewayClasses. The latter is similar to the "latest" release approach.

design/service-apis.md
stevesloka: I agree with you @danehans that the `GatewayClass` is a place to describe how a `Gateway` is configured, then the `Gateway` is an instantiation of a specific `GatewayClass`.

design/service-apis.md
stevesloka: It feels like there's too much logic applied with convention to the operator. We might need a version field in the spec to define what version of Contour you'd like to run for a specific instance of Contour that's managed by the operator. Even if that field could be set to `latest` if folks want truly the latest version or if they want to manage when the upgrade gets applied to their instances. A good use-case might be to upgrade the external version of Contour before upgrading the internal version. 

> The status of projectcontour.io/contour-operator/v1.11.0 GatewayClasses are updated to "Waiting" and associated Gateways are updated to "NoSuchGatewayClass". 

How did this happen? Did the operator change the GatewayClass somehow? It seems like if you upgraded the instance of the operator, and it wasn't programmed to look at a specific class, it would just ignore? 

> Optionally, the infrastructure provider updates the Contour resource referenced by the GatewayClass, or creates and associates a new Contour to the GatewayClass.

We need to be careful about how users need to manage their external services tied to Envoy, since in many cases, they are associated with DNS, etc.

youngnick:
I think that there's some confusion arisen here because the `spec.controller` field of GatewayClass is used for **two** things by Contour and the Contour Operator in our current design.

Firstly, the Contour Operator uses it to look at all GatewayClasses and build a Gateway for each one that's under its scope (however that's defined). I'd call this "Provisioning a GatewayClass" or something similar.

Secondly, Contour itself uses it to look for the Gateways it should be looking for as a starting point to find the *Route objects it cares about. I call this "Implementing the GatewayClass", as distinct from *provisioning it*.

From Contour's point of view, it doesn't care how the GatewayClass and Gateways get created, just that there are ones, and they correctly reference HTTPRoute or TLSRoute objects. Theoretically, it's possible to configure a GatewayClass and Gateway(s) that are usable by Contour, even without the Operator, so Contour *must* be able to handle this case (even if by "handle" we mean "throw an error and ignore that config").

So, it's okay, and not a conflict, for the Operator to say "Contours managed by the Operator will have a `spec.controller` of `projectcontour.io/contour-operator`". (But I think that building the version into that string doesn't feel like the right way to solve the upgrade problem.)

Contour itself will specify that the default for the GatewayClass `spec.controller` field is `projectcontour.io/contour`, and the operator will have the different value. This allows you to tell at a glance whether a particular Contour installation is Operator-managed or not (since we will still be supporting non-Operator installs).

@danehans, it seems like in the case that you're using the Service APIs, a Contour installation (including Envoys) will be created both by the `Contour` resource being present, and by the `Gateway` being present. I assumed that the Operator would *create* the Gateway resource to describe the Contour installation, but I suppose it doesn't really matter. I think we just need to be clear here about the difference in behavior for the operator when you're using the Service-APIs and when you're not.

Can you switch seamlessly between Service APIs and not using the same Contour object without redeploying? That seems like a nice feature, but I don't think it will be possible without really good coordination between Contour and the Operator, since Contour needs to know if it should be looking for GatewayClasses, and the Operator needs to create the Contour that needs to know about that.

I think there's a lot to cover here, and that we should not go too much further with either this design or projectcontour/contour#3213 until we ensure that the models are compatible. I think we should do that in this week's maintainer meeting (that is, the meeting on Wednesday 13th January Pacific time). @danehans, I'll ensure you have the invite to that.

design/service-apis.md
```diff
@@ -0,0 +1,384 @@
...
+    protocol: TCP
+   ...
+```
+
+__Note:__ Contour is responsible for constructing routing rules and forwarding traffic based on other `listeners`
+fields.
+
+TLS configuration must be included in the Gateway listener when "HTTPS" or "TLS" is specified for
+`protocol`. The `certificateRef` field is a reference to a Secret in the Gateway's namespace containing a TLS
+certificate and private key. The secret MUST contain the following keys and data:
```
youngnick, line 214: Maybe a note that the `kubernetes.io/tls` Secret type has these?

design/service-apis.md
```diff
@@ -0,0 +1,384 @@
...
+data:
+  tls.crt: <CERT>
+  tls.key: <KEY>
+```
+
+Gateway contains a `routes` field that specifies a schema for associating routes with the `listener` using Kubernetes
+selectors. A Route is a Service APIs resource capable of servicing a request and allows a user to expose a resource
+(i.e. Service) by an externally-reachable URL, proxy traffic to the resource, and terminate SSL/TLS connections.
+
+The following example will bind the Gateway to Routes that include label "app: example":
```
youngnick, line 251:
```suggestion
The following example will bind the Gateway to HTTPRoutes that include label "app: example":
```

design/service-apis.md
youngnick: Building version management into this fields smells a bit funny to me. Nothing I can concretely say is wrong, but it feels like a problem that should be solved elsewhere, not using string parsing of this field.

design/service-apis.md
```diff
@@ -0,0 +1,384 @@
...
+- Optionally, the Platform Provider updates the Contour resource referenced by the GatewayClass, or creates and
+  associates a new Contour to the GatewayClass. For example, a new configuration field is exposed through the Contour
+  resource in y.
+- The Platform Provider updates Gateways by setting GatewayClasses to `projectcontour.io/contour-operator/y`. This
+  triggers the operator to upgrade Contour/Envoy to y and optionally update the configuration.
+  GatewayClass|Gateway|xRoute status conditions are updated accordingly.
+
+## Implementation Details
+
+### Controller
```
youngnick, line 342: I think we need to settle on nomenclature here, between how we refer to the thing that's in `proejctcontour/contour` at the moment, and how we refer to the Operator. I suggest that this should be `Operator` and the repo is the Contour operator, and we call `projectcontour/contour` the Contour controller. That lets us avoid things like "Contour proper", or other slightly tortured phrases.

design/service-apis.md
```diff
@@ -0,0 +1,384 @@
...
+resources.
+
+### Test Plan
+
+- Add e2e, integration, and unit tests.
+- Create a CI job to run tests.
+
+### Open Questions
+
+- For versioned GatewayClass support, i.e. `controller: projectcontour.io/contour-operator/v1.12.0`, should Contour be
```
youngnick, line 373: I think we're better to punt more careful upgrades down the road a bit and say that when you upgrade the operator to a version, it will immediately upgrade all controllers under its management. That way we can work on getting the intial install working first, which is after all a prerequisite for being able to upgrade.

youngnick (commented):
I've added a few more comments ahead of our catchup about this tomorrow. I think that the key part here is to ensure that the underlying model we are using is in harmony, then the details will follow more easily.

danehans:
@youngnick thanks a bunch for taking the time to review the design spec.

> So, it's okay, and not a conflict, for the Operator to say "Contours managed by the Operator will have a spec.controller of projectcontour.io/contour-operator". (But I think that building the version into that string doesn't feel like the right way to solve the upgrade problem.)

I can remove the version content from the design spec and have the operator handle upgrades in the current manner, i.e. update the image tags in the operator's Deployment.

> Can you switch seamlessly between Service APIs and not using the same Contour object without redeploying?

In theory, a user should be able to remove `GatewayClassRef` or change the value to "None" (default) and the operator will manage the Contour without Service APIs. If this is not the desired, then we can implement safeguards to prevent this behavior.

design/service-apis.md
danehans: Good point, I'll update the doc and xref: https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets

design/service-apis.md
danehans: OK, I was trying to following the Service API terms, using "controller" instead of "operator".

danehans:
xref 1/13/21 maintainers meeting notes: https://docs.google.com/document/d/1vIXLcsqaZgtZGq4KyalKF6ecwOt-Hjt-sYeB9VzqApY/edit?usp=sharing

design/service-apis.md
```diff
@@ -0,0 +1,389 @@
...
+apiVersion: operator.projectcontour.io/v1alpha1
+kind: Contour
+metadata:
+  name: example-contour
+  namespace: contour-operator
+spec:
+  gatewayClassRef: example-gatewayclass
+  replicas: 2
+  namespace:
+    name: projectcontour
```
skriss, line 148: Since a `Contour` defines a namespace, does this mean that each `GatewayClass` can only be instantiated in a single namespace?

What happens if a user defines a `Gateway` for this `GatewayClass`, in a namespace other than `projectcontour`?

More generally, I wonder if there are, or will be, fields in the `Contour` CRD spec that won't make sense as pre-defined parameters for a `GatewayClass`, and if so, whether we'd want to use something other than a `Contour` CR to define `GatewayClass` config.

danehans: > Since a Contour defines a namespace, does this mean that each GatewayClass can only be instantiated in a single namespace?

@skriss GatewayClass is a cluster-scoped resource.

> What happens if a user defines a Gateway for this GatewayClass, in a namespace other than projectcontour?

@skriss currently, `gatewayclass.spec.parametersRef` should reference a cluster-scoped resource. This will be changing in the future to support cluster-scoped and namespaced-scoped resource references. We'll see if the parametersRef needs to be in the same namespace as Gateways of this GatewayClass. In the meantime, I suggest we don't impose such restriction. Conceivably, one persona can be responsible for managing cluster-wide proxy configs, i.e. GatewayClasses/Contours and a different persona be responsible for managing Gateways.

> More generally, I wonder if there are, or will be, fields in the Contour CRD spec that won't make sense as pre-defined parameters for a GatewayClass, and if so, whether we'd want to use something other than a Contour CR to define GatewayClass config.

I originally went in the direction by proposing the `ContourConfig` config resource for Service APIs. @jpeach suggested that we continue using the @Contour resource and I tend to agree with him. Until we have a solid use-case for not using the Contour CRD, I suggest we stick with it.

youngnick: With respect to having a Gateway-specific thing, I think it's fine to leave the `Contour` object reference to start with, we can change our minds to another thing later. I think that any other thing would be substantially similar to the `Contour`, so we'd probably end up needing to do an embedded struct or something anyway, to cover the similarities.

design/service-apis.md
stevesloka: It feels to like the `GatewayClass` should tell us this information (e.g. the operator is managing). The operator will react to any `Gateways` being created and deploy the Contour infrastructure, but at that point it's hands-off. Contur the controller then will watch for the specific Gateway it's responsible to manage (e.g. IngressClass) and process related `xRoute` resources. 

I think we can have quick discussion on this next time we're all together and hammer out an answer.

danehans:
@youngnick @stevesloka @skriss is anything else needed to merge this design spec?

stevesloka (approved):
/lgtm 🎉

skriss (approved):
I think we're in good shape here, can sort out any further issues in implementation.

Issue #141, opened 2020-12-02
Add Migration Support


danehans:
**Please describe the problem you have**
Many users are manually managing Contour, i.e. getting started guide. How can users migrate a manual deployment to an operator-managed deployment.

An e2e test should be added to test migration from the example rendered manifests to the example `Contour`.

/cc @Miciah

danehans:
__Test Case 1: Migrate from the rendered manifests__

Steps
1. Run a kind cluster:
```make load-cluster```
2. Install Contour using the rendered manifests:
```kubectl apply -f https://projectcontour.io/quickstart/contour.yaml```
3. Run a sample app:
```https://projectcontour.io/examples/kuard.yaml```
4. Test e2e functionality:
```curl -s -o /dev/null -w "%{http_code}" http://local.projectcontour.io```
5. Run the operator:
```kubectl apply -f https://projectcontour.io/quickstart/operator.yaml```
6. Create a Contour custom resource with default settings:
```kubectl apply -f https://projectcontour.io/quickstart/contour-custom-resource.yaml```
 
Although the example Contour custom resource and the Contour example rendered manifests are very similar, they are __NOT__ identical. For example, the Contour Operator will fail to manage the Contour Deployment and Envoy DaemonSet resources due to `selector` differences. The rendered manifest uses the "app: contour" while Contour Operator applies a label that identifies the owning Contour resource, i.e. "contour.operator.projectcontour.io/deployment-contour: contour-example". Since selectors are immutable, the following error occurs when the operator tries to manage the existing Deployment and DaemonSet created by the rendered manifests:

```
[...]

PR #173, opened 2021-01-13
Updates Deployment/DaemonSet Selectors

danehans:
Previously, the Deployment and DaemonSet used selectors that reference the associated `Contour`. This prevents users from migrating existing Contour deployments to operator-managed Contour deployments. This PR updates the selectors to use the `app: contour|envoy` scheme used by Contour's example manifests. This PR should be reverted when [Issue 1821](https://github.com/projectcontour/contour/issues/1821) is fixed.

xref: https://github.com/projectcontour/contour-operator/issues/141


stevesloka:
@danehans I know this is Alpha, but this would break any existing installs or just update the selectors automatically to what's in this PR when the new operator is deployed?

danehans:
> @danehans I know this is Alpha, but this would break any existing installs or just update the selectors automatically to what's in this PR when the new operator is deployed?

@stevesloka commit `a57d786` adds support for the label transition by recreating the Deployment and DaemonSet resources.

youngnick (approved):
Overall this LGTM, given that the operator is experimental, it's okay for there to be a routing outage on change. But we definitely to make sure that people understand that upgrading the operator is possibly outage-causing (for now).

stevesloka (approved):

PR #174, opened 2021-01-13
Updates Examples Test Timeout

danehans:
In some instances, it takes longer than 50-seconds for an instance of Contour to become available and begin serving traffic. This PR increases the number of test curl's to 100 to prevent test flakes.


stevesloka (approved):
/lgtm

PR #177, opened 2021-01-18
Adds Contour Owner Label to All Resources

danehans:
Previously, the operator would manage resources that matched a `Contour` even when the resources were provisioned by other mechanisms, i.e. `kubectl apply -f https://projectcontour.io/quickstart/contour.yaml`. This PR:
- Ensures only resources that include the `Contour` owner label are managed by the operator.
- Decomposes RBAC resource management into separate `<resource>.go` files.
- Adds unit tests for RBAC and Namespace resources.
- Updates the RBAC and Namespace resource management model from Create/Delete to Create/Update/Delete, providing management consistency other resources, i.e. DaemonSet.
- Updates the Job naming scheme to include the Contour version.

Requires: https://github.com/projectcontour/contour-operator/pull/173
Partially fixes: https://github.com/projectcontour/contour-operator/issues/141

### DIFF
--- a/internal/object/object.go
+++ b/internal/object/object.go
@@ -11,9 +11,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package object
+package util
 
 import (
+	"strings"
+
 	operatorv1alpha1 "github.com/projectcontour/contour-operator/api/v1alpha1"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -198,4 +200,14 @@ func NewIngress(name, ns, backendName string, backendPort int) *networkingv1.Ing
 			},
 		},
 	}
+}
+
+// TagFromImage returns the tag from the provided image or an
+// empty string if the image does not contain a tag.
+func TagFromImage(image string) string {
+	if strings.Contains(image, ":") {
+		parsed := strings.Split(image, ":")
+		return parsed[1]
+	}
+	return ""
 }
--- a/internal/operator/controller/contour/cluster_role.go
+++ b/internal/operator/controller/contour/cluster_role.go
@@ -1,0 +1,167 @@
+// Copyright Project Contour Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package contour
+
+import (
+	"context"
+	"fmt"
+
+	operatorv1alpha1 "github.com/projectcontour/contour-operator/api/v1alpha1"
+	"github.com/projectcontour/contour-operator/internal/equality"
+	objutil "github.com/projectcontour/contour-operator/internal/object"
+
+	projectcontourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// ensureClusterRole ensures a ClusterRole resource exists with the provided name
+// and contour namespace/name for the owning contour labels.
+func (r *reconciler) ensureClusterRole(ctx context.Context, name string, contour *operatorv1alpha1.Contour) (*rbacv1.ClusterRole, error) {
+	desired := desiredClusterRole(name, contour)
+	current, err := r.currentClusterRole(ctx, name)
+	if err != nil {
+		if errors.IsNotFound(err) {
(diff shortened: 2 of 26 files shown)


controller/contour/controller.go
```diff
@@ -282,3 +286,27 @@ func contourFinalized(contour *operatorv1alpha1.Contour) bool {
...
+	labels := obj.(metav1.Object).GetLabels()
+	nameFound := false
+	nsFound := false
+	if labels == nil {
+		return false
+	}
+	for l, v := range labels {
+		switch {
+		case nameFound && nsFound:
+			break
```
stevesloka, line 301: nit: Could just return here since both name & namespace are found in this case.

controller/contour/deployment_test.go
```diff
@@ -16,6 +16,8 @@ package contour
 import (
 	"testing"
 
+	oputil "github.com/projectcontour/contour-operator/util"
+
```
stevesloka, line 20: nit: remove space

stevesloka (commented):
Overall this looks good! I think there are some conflicts to clean up then can do a final pass over it all.

controller/contour/deployment_test.go
danehans: @stevesloka imports for the project have been organized as such:

1. standard libs
2. local libs
3. non-std libs

danehans:
@stevesloka commit `a6d4382` resolves your most recent comments.

controller/contour/cluster_role.go
```diff
@@ -0,0 +1,169 @@
...
+	r.Log.Info("created cluster role", "name", cr.Name)
+
+	return cr, nil
+}
+
+// updateClusterRoleIfNeeded updates a ClusterRole resource if current does not match desired,
+// using contour to verify the existence of owner labels.
+func (r *Reconciler) updateClusterRoleIfNeeded(ctx context.Context, contour *operatorv1alpha1.Contour, current, desired *rbacv1.ClusterRole) (*rbacv1.ClusterRole, error) {
+	exist := doOwnerLabelsExist(current, contour)
+	if !exist {
```
skriss, line 153: nit: I'd usually just write `if !doOwnerLabelsExist(...)` and drop the variable declaration since it's only used in one place. Not a big deal, just a little simpler IMHO. This applies to the other calls to `doOwnerLabelsExist` as well.

controller/contour/controller.go
```diff
@@ -286,3 +286,28 @@ func contourFinalized(contour *operatorv1alpha1.Contour) bool {
 
 	return false
 }
+
+// doOwnerLabelsExist returns true if obj contains Contour owner labels.
+func doOwnerLabelsExist(obj runtime.Object, contour *operatorv1alpha1.Contour) bool {
+	labels := obj.(metav1.Object).GetLabels()
```
skriss, line 292: can this function take a `metav1.Object` directly instead of a `runtime.Object`, so we don't have to type-assert here?

skriss (changes requested):
Just a couple small comments.

For future reference, I think it'd be easier for reviewers if you keep refactoring in 1+ separate PRs - there are several different changes going on here which makes it more difficult to do a good review on.

skriss (approved):
LGTM

internal/object/object.go
```diff
@@ -199,3 +201,9 @@ func NewIngress(name, ns, backendName string, backendPort int) *networkingv1.Ing
 		},
 	}
 }
+
+// TagFromImage returns the tag from the provided image.
+func TagFromImage(image string) string {
+	parsed := strings.Split(image, ":")
```
stevesloka, line 207: This works as long as we don't pass something like: `projectcontour/contour` which I think we'll be ok with since it's internally managed.

internal/operator/controller/contour/cluster_role.go
```diff
@@ -0,0 +1,168 @@
...
+	crd := rbacv1.PolicyRule{
+		Verbs:     []string{"list"},
+		APIGroups: groupExt,
+		Resources: []string{"customresourcedefinitions"},
+	}
+	svcAPI := rbacv1.PolicyRule{
+		Verbs:     verbGLW,
+		APIGroups: groupNet,
+		// TODO [danehans]: Update roles when v1alpha1 Service APIs are released.
+		Resources: []string{"gatewayclasses", "gateways", "httproutes", "tcproutes"},
```
stevesloka, line 96: Does the operator care about httproute, tlsroutes, & tcproutes?

stevesloka: Or should it?

stevesloka, line 96: Should we add BackendPolicy to this?

danehans: ack

internal/object/object.go
danehans: @stevesloka I updated the func to return  "" if the provided image does not contain `:`.

danehans:
@stevesloka commit `e05a27c` resolves your comments above.

stevesloka (approved):
/lgtm

internal/object/object.go
```diff
@@ -199,3 +201,13 @@ func NewIngress(name, ns, backendName string, backendPort int) *networkingv1.Ing
...
 }
+
+// TagFromImage returns the tag from the provided image or an
+// empty string if the image does not contain a tag.
+func TagFromImage(image string) string {
+	if strings.Contains(image, ":") {
+		parsed := strings.Split(image, ":")
+		return parsed[1]
+	}
+	return ""
```
stevesloka, line 212: Well should this return `image` instead of ""?

danehans: @stevesloka I don't think so. The purpose of the func is to return the tag from the provided image. If the image does not contain a tag, I would expect an empty tag, i.e. "" or an error.

PR #182, opened 2021-01-27
Implements GatewayClassRef in Support of Gateway API

danehans:
Implements `GatewayClassRef` in support of Gateway API. The contour controller will only instantiate a Contour environment if "None" is defined for `GatewayClassRef` or if the field is undefined (defaults to "None"). The "None" value signifies that a GatewayClass is not responsible for managing the Contour custom resource.

Requires: https://github.com/projectcontour/contour-operator/pull/180
xref: https://github.com/projectcontour/contour-operator/pull/171

### DIFF
--- a/api/v1alpha1/contour_types.go
+++ b/api/v1alpha1/contour_types.go
@@ -75,6 +75,12 @@ type ContourSpec struct {
 	//
 	// +kubebuilder:default={envoy: {type: LoadBalancerService, containerPorts: {{name: http, portNumber: 8080}, {name: https, portNumber: 8443}}}}
 	NetworkPublishing NetworkPublishing `json:"networkPublishing,omitempty"`
+
+	// GatewayClassRef is a reference to a GatewayClass name used for
+	// managing a Contour.
+	//
+	// +kubebuilder:default=None
+	GatewayClassRef string `json:"gatewayClassRef,omitempty"`
 }
 
 // NamespaceSpec defines the schema of a Contour namespace.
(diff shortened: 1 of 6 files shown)


danehans:
Note that the soon-to-be implemented gateway controller will be responsible for instantiating a Contour environment when a valid GatewayClassRef is provided.

go.mod
```diff
@@ -17,4 +17,5 @@ require (
 	k8s.io/utils v0.0.0-20200729134348-d5654de09c73
 	sigs.k8s.io/controller-runtime v0.6.3
 	sigs.k8s.io/controller-tools v0.4.0
+	sigs.k8s.io/service-apis v0.1.1-0.20210114050902-2b2bf1b8fdf5
```
danehans, line 20: @stevesloka FYI I had to bump service apis to this version to get a fix for `gatewayclass.status`.

stevesloka: Ahh ok cool! We'll have to bump Contour to match the same.

youngnick (approved):
This LGTM for the happy path, but I'm curious as to what would happen for a less-happy case (to sort out at some point, doesn't block this PR):
- what happens when someone deletes the Contour that's in use by a Gateway? So a GatewayClass refers to a Contour, and there's a Gateway that refers to that GatewayClass. What happens when that Contour is removed? Is it reconfigured back to defaults? Removed then reprovisioned as default?

danehans:
> what happens when someone deletes the Contour that's in use by a Gateway? So a GatewayClass refers to a Contour, and there's a Gateway that refers to that GatewayClass. What happens when that Contour is removed? Is it reconfigured back to defaults? Removed then reprovisioned as default?

@youngnick I'm working through ^ use cases in https://github.com/danehans/contour-operator/tree/wip_svc_apis_impl.

youngnick:
To keep this moving, I'm going to merge with only my approval.

PR #190, opened 2021-02-03
Fixes Envoy Service Equality Check

danehans:
Fixes an issue where the incorrect equality check was being used for the two supported network publishing types.


skriss (approved):
LGTM, oof. Would be a good area to add to a follow-up for improving test coverage.

danehans:
> Would be a good area to add to a follow-up for improving test coverage.

@skriss I agree, but the bug is only exposed when using the LoadBalancerService network publishing type. How does Contour handle testing against a cloud LB environment? MetalLB can be used to simulate a service of type LoadBalancer.

youngnick (commented):
LGTM, nice find.

PR #193, opened 2021-02-03
Updates RABC for IngressClass and Service APIs

danehans:
- Updates RBAC for IngressClass due to https://github.com/projectcontour/contour/pull/3266.
- Updates RBAC for Service APIs.
- Bumps deps.


internal/operator/controller/contour/controller.go
```diff
@@ -63,7 +63,7 @@ type reconciler struct {
 // +kubebuilder:rbac:groups="",resources=namespaces;secrets;serviceaccounts;services,verbs=get;list;watch;delete;create;update
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;delete;create;update
 // +kubebuilder:rbac:groups="",resources=endpoints,verbs=get;list;watch
-// +kubebuilder:rbac:groups=networking.k8s.io,resources=gatewayclasses;gateways;httproutes;tcproutes;ingresses;backendpolicies,verbs=get;list;watch
+// +kubebuilder:rbac:groups=networking.k8s.io,resources=gatewayclasses;gateways;httproutes;tcproutes;ingresses;ingressclasses;backendpolicies,verbs=get;list;watch
```
stevesloka, line 66: As an aside, should the service-apis group be `networking.x-k8s.io` and not `networking.k8s.io`? This is what I have in Contour which puts the `ingresses` & `ingressclasses` into different groups.

stevesloka (approved):

stevesloka (changes requested):
Just one question on the service-apis group.

internal/operator/controller/contour/controller.go
youngnick: The apigroup is definitely `networking.x-k8s.io` in the upstream code for the service-apis, but should be `networking.k8s.io` for all the Ingress and IngressClass stuff.

danehans: ack

danehans:
@stevesloka @youngnick commit `4b7b2bd` updates the PR based on the https://github.com/projectcontour/contour-operator/pull/193#discussion_r569746914. The commit includes the following:

1. Updates the Service APIs group name.
2. Updates Service APIs RBAC to only include Gateway and GatewayClass resources. The Contour ingress controller will be responsible for the other API types, i.e. HTTPRoute.
3. Bumped deps for importing Service APIs. __Note:__ I used a commit ref instead of a versioned release to a get a fix for `gatewayclass.status`. (required for https://github.com/projectcontour/contour-operator/pull/182).

youngnick (approved):
LGTM, nice work

config/rbac/role.yaml
```diff
@@ -99,6 +95,24 @@ rules:
...
+  - gateways
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - networking.x-k8s.io
+  resources:
+  - gatewayclasses/status
+  - gateways/status
```
stevesloka, line 111: Is the Operator going to write status to the `Gateway`? Or is that the role of Contour?

danehans: @stevesloka the operator will be responsible for managing Contour, GatewayClass, and Gateway objects and therefore be responsible for writing status to them. Contour will be responsible for app-level Service API object, i.e. HTTPRoute.

youngnick: They will both end up writing some status; the operator will need to list the address(es) the Gateway has ended up with, and the controller will need to update with any errors in handling Routes. I don't think there's any field crossover though.

stevesloka: got it, thanks for clarifying @danehans! 

That makes sense, but what would happen in the event we deploy Contour without the Operator? Who is responsible for updating the same info? 

To me, it feels like Contour should be monitoring the service which backs Envoy and watching the endpoints that get assigned and updating (which is what happens today) and the operator is responsible for setting up the proper bits to request those IPs depending on its environment (i.e. cloud, bare metal, etc).

danehans: > That makes sense, but what would happen in the event we deploy Contour without the Operator? Who is responsible for updating the same info?

The controller used to manage Service APIs infrastructure-level resources, i.e. Gateway/GatewayClass, is responsible for managing status of those objects. If the operator is not being used, some other controller must actively reconcile GatewayClass and Gateway resources. IMO it doesn't make sense for Contour to manage these resources since it's an app-level controller.

Status for a Gateway consists of:
- __Addresses:__ IP addresses used by all Gateway listeners. The operator will use the IP from the Envoy Service for this field.
- __Conditions:__ Standard Kubernetes status conditions. The operator will apply the "Scheduled" condition when instantiating a Contour environment and update to "Ready" when the Deployment/DaemonSet are ready.
- __Listeners:__ The status associated for each Gateway listener. The operator will populate port and protocol fields from the Envoy Service for this field. The operator will copy `gateway.spec.listeners[].hostname` to this status field. Listener conditions will be equivalent to gateway conditions unless Contour manages this field and provides more specific status.

stevesloka: Let's chat about it in Contour Maintainer's meeting, but I think if you deployed Contour with say Helm or example yaml files, you shouldn't need another controller just to manage state on the Gateway (e.g. Address/Listeners). Contour should be responsible for them like it does already today to update the IngressStatus information on Ingress/HTTPProxy resources.

Issue #191, opened 2021-02-03
LoadBalancerService Network Publishing e2e Test Coverage


danehans:
**Please describe the problem you have**
Expand e2e test coverage to support the LoadBalancerService network publishing type. Currently, e2e tests only cover the NodePortService type because [kind](https://kind.sigs.k8s.io/) is being used to expose the Envoy service via host ports:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
  extraPortMappings:
  - containerPort: 30080
    hostPort: 80
    listenAddress: "0.0.0.0"
  - containerPort: 30443
    hostPort: 443
    listenAddress: "0.0.0.0"
```

The following references indicate that kind can be used in conjunction with [MetalLB](https://metallb.universe.tf/) to support LoadBalancer services:
- https://mauilion.dev/posts/kind-metallb/
- https://www.thehumblelab.com/kind-and-metallb-on-mac/

xref: https://github.com/projectcontour/contour-operator/pull/190

sunjayBhatia:
Contour uses a load balancer service in integration tests without any extra setup besides telling the integration tester to connect to local host: https://github.com/projectcontour/contour/blob/9facf49dd1a29227d87881ba6f091c9f133260c9/examples/contour/02-service-envoy.yaml#L26 and https://github.com/projectcontour/contour/blob/9facf49dd1a29227d87881ba6f091c9f133260c9/_integration/testsuite/run-test-case.sh#L25

youngnick:
This sounds interesting! Hopefully it can be done without impacting the time for the integration tests (Contour's integration tests are already not very fast).

stevesloka:
I was actually playing with this today for upgrade testing. The knative folks have metallb running in kind for their net-contour integration tests: https://github.com/knative-sandbox/net-contour/blob/master/.github/workflows/kind-e2e.yaml#L123

This will help with sending traffic to nodes while doing upgrade testing (different issue specifically). 

//cc @mattmoor

mattmoor:
Yeah it works great, I have been using it with all of the mink testing as well

danehans:
@sunjayBhatia thanks for the references above. I was able to test simulating a LB service using "127.0.0.1" as the LB IP and it worked. I'll be submitting a PR to update the operator's e2e tests accordingly.

PR #195, opened 2021-02-03
Updates e2e Tests

danehans:
Updates the e2e tests to:
1. Test the "LoadBalancerService" network publishing type. __Note:__ This is the default type of a Contour.
2. Adds a step to each e2e test that parses the operator's logs for error messages.

Requires: https://github.com/projectcontour/contour-operator/pull/193
Fixes: https://github.com/projectcontour/contour-operator/issues/191

### DIFF
--- a/internal/parse/parse.go
+++ b/internal/parse/parse.go
@@ -20,6 +20,8 @@ import (
 	_ "crypto/sha256"
 	_ "crypto/sha512"
 	"fmt"
+	"os/exec"
+	"strings"
 
 	"github.com/docker/distribution/reference"
 )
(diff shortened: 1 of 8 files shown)


internal/parse/parse.go
```diff
@@ -34,3 +36,47 @@ func Image(s string) error {
...
 }
+
+// DeploymentLogsForString parses the container logs of the specified ns/name
+// deployment, returning true if the string was found.
+func DeploymentLogsForString(ns, name, container, expectedString string) (bool, error) {
+	cmdPath, err := exec.LookPath("kubectl")
+	if err != nil {
+		return false, err
+	}
+	slashName := fmt.Sprintf("deployment/%s", name)
```
youngnick, line 47: Are we not doing this with client-go because of dependency things? Seems like it would probably be better to use client-go if we could, to avoid shelling out.

youngnick (changes requested):
This LGTM with the exception of one query about the getting the logs. I think we may be managing the risk of having a function that runs an arbitrary command with arbitrary arguments, but that is still a bit scary, so we should either not do that, or note what we're doing to minimise the risks.

internal/parse/parse.go
danehans: @youngnick I created https://github.com/projectcontour/contour-operator/issues/200 to track transitioning from the controller-runtime client to client-go. I would like to stick with the current approach, since this PR does help prevent regressions and it will allow me to focus on service apis.

internal/parse/parse.go
```diff
@@ -34,3 +36,47 @@ func Image(s string) error {
...
+	if strings.Contains(result, expectedString) {
+		return true, nil
+	}
+	return false, nil
+}
+
+// runCmd runs command cmd with arguments args and returns the output
+// of the command or an error.
+func runCmd(cmd string, args []string) (string, error) {
+	execCmd := exec.Command(cmd, args...)
```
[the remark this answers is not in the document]
danehans: I looked through the `exec.Command` code and I don't see any mechanisms to prevent an arg that is prefixed by `/` or `.`. I've updated `runCmd()` based on your feedback.

internal/parse/parse.go
```diff
@@ -34,3 +36,52 @@ func Image(s string) error {
 
 	return nil
 }
+
+// DeploymentLogsForString parses the container logs of the specified ns/name
+// deployment, returning true if the string was found.
```
youngnick, line 41: The changes to this look good, could you add a TODO here, linking to #200, so that we can have a full bidirectional reminder? :)

youngnick (approved):
LGTM with one small doc request.

danehans:
@youngnick commit `d30014b` resolves your feedback in https://github.com/projectcontour/contour-operator/pull/195#discussion_r575920415.

PR #196, opened 2021-02-05
Bump golang to 1.15.8

sunjayBhatia:
See release notes: https://golang.org/doc/devel/release.html#go1.15

Set Dockerfile to use an explicit version of Go so we can explicitly control the dependency ourselves.

See also https://github.com/projectcontour/contour/pull/3325


skriss (approved):
LGTM assuming CI's green

Issue #28, opened 2020-09-17
Remove Logging for "Skipped..." Operations


danehans:
**Please describe the problem you have**
The operator logs many events when an operation is being skipped. Although this is nice for early development, it should be removed to reduce noise as the operator matures.

xref: https://github.com/projectcontour/contour-operator/pull/21#discussion_r489086428

/cc @Miciah

Miciah:
The logging is useful during development, so we won't remove this for v1alpha1.

Alternatively, we might want to keep this logging even after alpha but set the logging level to "debug" (or equivalent).

Issue #106, opened 2020-11-06
Add Gateway API Support


danehans:
**Please describe the problem you have**
Add [Gateway API](https://github.com/kubernetes-sigs/gateway-api) support, specifically GatewayClass and Gateway Support.

xref [design doc](https://github.com/projectcontour/contour-operator/blob/main/design).

/cc @Miciah @jpeach @youngnick

danehans:
xref: https://github.com/projectcontour/contour-operator/issues/111

PR #204, opened 2021-02-16
Contour Controller and Logging Refactor

danehans:
- Refactors the `ensureX()` functions for `Contour` child resources, i.e. Contour ConfigMap. This will provide a shared location for future controllers, i.e. Gateway, to call `ensureX()` functions.
- Removes logging for "Skipped..." operations.

Fixes https://github.com/projectcontour/contour-operator/issues/28


internal/objects/rolebinding/role_binding.go
```diff
@@ -11,39 +11,55 @@
...
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// ensureRoleBinding ensures a RoleBinding resource exists with the provided
+// New makes a RoleBinding object using the provided
+// ns/name for the object's namespace and name.
+func New(ns, name string) *rbacv1.RoleBinding {
```
stevesloka, line 34: This doesn't need exported.

internal/objects/object.go
```diff
@@ -0,0 +1,68 @@
...
+		RunAsUser:    &user,
+		RunAsGroup:   &group,
+		RunAsNonRoot: &nonRoot,
+	}
+}
+
+// NewIngress makes an Ingress using the provided ns/name for the
+// object's namespace/name and backendName/backendPort as the name
+// and port of the backend Service.
+func NewIngress(name, ns, backendName string, backendPort int) *networkingv1.Ingress {
```
stevesloka, line 40: Move this to the e2e testing package since it's only used for testing.

internal/objects/clusterrolebinding/cluster_role_binding.go
```diff
@@ -11,50 +11,64 @@
...
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// ensureClusterRoleBinding ensures a ClusterRoleBinding resource with the provided
+// New makes a ClusterRoleBinding object using
+// the provided name for the object's name.
+func New(name string) *rbacv1.ClusterRoleBinding {
```
stevesloka, line 34: This doesn't need exported.

internal/objects/contour/contour.go
```diff
@@ -0,0 +1,123 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+// OwnerLabelsExist returns true if obj contains Contour owner labels.
+func OwnerLabelsExist(obj metav1.Object, contour *operatorv1alpha1.Contour) bool {
```
stevesloka, line 60: Refactor this into instance methods to make it cleaner maybe? 

e.g.
```
func (c *operatorv1alpha1.Contour)OwnerLabelsExist(obj metav1.Object) bool {}
```

internal/objects/contour/contour.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	if len(contours.Items) == 0 || len(contours.Items) == 1 && contours.Items[0].Name == contour.Name {
+		return false, nil, nil
+	}
+	return true, contours, nil
+}
+
+// OtherContoursExistInSpecNs lists Contour objects in the same spec.namespace.name as contour,
+// returning true if any exist.
+func OtherContoursExistInSpecNs(ctx context.Context, cli client.Client, contour *operatorv1alpha1.Contour) (bool, error) {
```
stevesloka, line 99: This could refactor into an instance method as well.

internal/objects/contour/contour.go
```diff
@@ -0,0 +1,123 @@
...
+				return true, nil
+			}
+		}
+	}
+	return false, nil
+}
+
+// OwningSelector returns a label selector using "contour.operator.projectcontour.io/owning-contour-name"
+// and "contour.operator.projectcontour.io/owning-contour-namespace" labels.
+func OwningSelector(contour *operatorv1alpha1.Contour) *metav1.LabelSelector {
```
stevesloka, line 116: This could refactor into an instance method as well.

stevesloka (changes requested):
I think breaking up the bits into packages is nice, since they will grow, so that pattern is great! I think there are a bunch of places that could be refactored into instance methods. I've commented on some of them but didn't want to do each since it's overkill and also get your thoughts on that idea.

danehans:
xref: https://github.com/projectcontour/contour-operator/issues/106

internal/objects/contour/contour.go
danehans: I cannot create an instance method on a non-local type. I need to wrap the external type and do the same for other pkgs that call OwnerLabelsExist(). Thoughts?

danehans:
@stevesloka other than turning `OwnerLabelsExist()` into a method, commit `5314158` resolves your feedback.

internal/objects/contour/contour.go
danehans: Same response as https://github.com/projectcontour/contour-operator/pull/204#discussion_r578085255.

internal/objects/contour/contour.go
danehans: Same response as https://github.com/projectcontour/contour-operator/pull/204#discussion_r578085255.

internal/objects/contour/contour.go
youngnick: I think we should either just do this, or put these functions in the exported API. If anyone else is doing things with Contours, they'll be handy anyway.

internal/objects/contour/validate.go
```diff
@@ -20,8 +20,8 @@ import (
 	"github.com/projectcontour/contour-operator/internal/slice"
 )
 
-// validateContour returns true if contour is valid.
-func validateContour(contour *operatorv1alpha1.Contour) error {
+// ValidateContour returns true if contour is valid.
+func ValidateContour(contour *operatorv1alpha1.Contour) error {
```
youngnick, line 24: I think there is an argument to be made that we put these helpers *in* the actual Contour object API specification. Yes, that means that we will have a GA behavior guarantee, but it also puts the *enforcement* of the API spec right next to the *definition* of the API spec.

youngnick: I don't see that the Contour operator will ever do anything other than configure Contour and Envoy installs, so we don't need to worry about exposing that we use a Contour controller and an Envoy install.

internal/objects/daemonset/daemonset.go
```diff
@@ -369,49 +359,51 @@ func DesiredDaemonSet(contour *operatorv1alpha1.Contour, contourImage, envoyImag
 	return ds
 }
 
-// currentDaemonSet returns the current DaemonSet resource for the provided contour.
-func (r *reconciler) currentDaemonSet(ctx context.Context, contour *operatorv1alpha1.Contour) (*appsv1.DaemonSet, error) {
+// urrentDaemonSet returns the current DaemonSet resource for the provided contour.
```
youngnick, line 362:
```suggestion
// currentDaemonSet returns the current DaemonSet resource for the provided contour.
```

internal/objects/rbac.go
```diff
@@ -11,77 +11,83 @@
...
 )
 
-// ensureRBAC ensures all the necessary RBAC resources exist for the
+// EnsureRBAC ensures all the necessary RBAC resources exist for the
 // provided contour.
-func (r *reconciler) ensureRBAC(ctx context.Context, contour *operatorv1alpha1.Contour) error {
+func EnsureRBAC(ctx context.Context, cli client.Client, contour *operatorv1alpha1.Contour) error {
 	ns := contour.Spec.Namespace.Name
-	names := []string{contourRbacName, envoyRbacName, certGenRbacName}
+	names := []string{ContourRbacName, EnvoyRbacName, CertGenRbacName}
```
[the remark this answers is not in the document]
youngnick: I understood that the operator would partition Contour installs per-namespace.

I think doing that is a better idea, as it allows for more granular RBAC etc. If you care enough about what each Contour install is doing, you should care enough to split them apart into separate namespaces.

youngnick (approved):
I think this is fine as it is, but if we want to make those validation helpers more usable, we should put them in the public API, as methods on the actual Contour object.

We've done that a bit with HTTPProxy, but I think it's a better pattern (even given the downside that you then lock in those functions if the API is GA).

internal/objects/contour/contour.go
stevesloka: Contour put a bunch of helpers like that in the API (https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/helpers.go). My thought is just that it's cleaner to say `contour.OwnerLabelsExist(obj)` vs passing around the `*operatorv1alpha1.Contour` each time. 

Could be a future refactor as well if needed.

pkg/validation/validation_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright Project Contour Authors
+// Copyright Project ValidateContour Authors
```
stevesloka, line 1: s/ValidateContour/Contour

pkg/validation/validation.go
```diff
@@ -11,22 +11,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package contour
+package validation
```
stevesloka, line 14: Does this need to be public? Why not leave in `internal`?

danehans: @stevesloka I put the validation funcs in a public pkg due to https://github.com/projectcontour/contour-operator/pull/204#discussion_r578122093. I can move to internal if you prefer.

internal/objects/contour/contour.go
stevesloka: Probably better the way it is than public in the api.

pkg/validation/validation.go
youngnick: I'm not strongly wedded to having them public, just seems like it would be useful for anyone actually consuming Contour objects to be able to use the standard validation rules rather than having them private.

PR #208, opened 2021-02-17
Renames Service APIs Design Spec

danehans:
Renames Service APIs design spec to -> Gateway API.


youngnick (approved):
LGTM, I should probably do this for the Controller one too.

skriss (approved):

Issue #159, opened 2020-12-10
Extend NetworkPublishing API


danehans:
Extend the NetworkPublishing API to support the following use cases:
- Set the `awsLbBackendProtoAnnotation: tcp` annotation only if the provider of `LoadBalancerServicePublishingType` is AWS.
- Add support to `NodePortServicePublishingType` to specify: 1) auto (default) or manual port allocation. If manual, provide fields to specify the desired http/https nodePorts (30080 default for http and 30443 default for https).

danehans:
xref: https://github.com/projectcontour/contour-operator/pull/151
xref: https://github.com/projectcontour/contour-operator/pull/152

stevesloka:
> Support to NodePortServicePublishingType to specify: 1) auto (default) or manual port allocation. If manual, provide fields to specify the desired http/https nodePorts (30080 default for http and 30443 default for https).

This implies the `default` by choosing option #1 is ports `30080` & `30443`. Just want to confirm that's the thought, in my head, `auto` implies that k8s is specifying the ports which is the default unless you specify the ports (e.g. `manual`).

danehans:
Fixed by:
- https://github.com/projectcontour/contour-operator/pull/367 (proxy proto support)
- #151 and #152 for NodePortService support.

PR #209, opened 2021-02-18
examples: Default example to use LoadBalancerService type 

stevesloka:
Defaults the example Contour config to use a LoadBalancer service to match the docs provided on the Contour quickstart.

Fixes https://github.com/projectcontour/contour/issues/3319

Signed-off-by: Steve Sloka <slokas@vmware.com>

### DIFF
--- a/examples/contour/contour.yaml
+++ b/examples/contour/contour.yaml
@@ -5,4 +5,4 @@ metadata:
 spec:
   networkPublishing:
     envoy:
-      type: NodePortService
+      type: LoadBalancerService


danehans:
@stevesloka we will need to support passing an IP, specifically 127.0.0.1, to the Envoy Service for an LB service to work on kind. This is why the example and CI use a "NodePortService". We will eventually need this to support specifying a `gateway.spec.addresses[]`. Maybe we should remove the `test-example-linux` job until the example can support all default fields of a Contour. Thoughts?

stevesloka:
A LoadBalancer service gives you a NodePort automatically, so if we switch to this all the examples will line up.

stevesloka:
I see why the test is failing, but not sure why we're hard coding the NodePort. Isn't there a chance that you'll conflict? Also, if you deployed a second instance it for sure would conflict in the same cluster.

danehans:
> A LoadBalancer service gives you a NodePort automatically, so if we switch to this all the examples will line up.

True, but I don't think we want to expose managing nodePorts for a "LoadBalancerService" type.

> I see why the test is failing, but not sure why we're hard coding the NodePort.

@stevesloka the node ports are being hard-coded due to the kind config:
```
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
  extraPortMappings:
  - containerPort: 30080
    hostPort: 80
    listenAddress: "0.0.0.0"
  - containerPort: 30443
    hostPort: 443
    listenAddress: "0.0.0.0"
```

> Isn't there a chance that you'll conflict? Also, if you deployed a second instance it for sure would conflict in the same cluster.

The [NetworkPublishing](https://github.com/projectcontour/contour-operator/blob/main/api/v1alpha1/contour_types.go#L128-L182) API does not currently expose port numbers for the `NodePortService` type. The plan is to add support as identified by https://github.com/projectcontour/contour-operator/issues/159.

youngnick:
It seems like the problem here is that the defaults are currently set to make it easier to test locally - that is, they're defaults for `kind`. @stevesloka is opening this PR to change the defaults so that they will match the Getting Started documentation in the event that you're installing using the Operator on a cloud or similar.

I think the larger question to answer for this PR is: which default is the right one? If we don't default to kind, then we will need to update the testing runs to ensure that we're testing the right things.

I think there are two options:
- accept this PR, change the default to Service of Type Loadbalancer, and update the testing, which may require us adding more to NetworkPublishing to allow it to describe a Kind deployment, as opposed to a generic cloud one.
- Change the getting started documentation to indicate that you get a Nodeport and no upstream load balancer unless you add NetworkPublishing bits to your Contour object.

stevesloka:
I agree, let's just leave this as-is and update the docs until the endpoint bits are ready.

Issue #213, opened 2021-02-23
Finalize GatewayClass and Contour when using Gateway API


danehans:
**Please describe the problem you have**
Currently, the associated GatewayClass and Contour of a Gateway is not being finalized. Both should be finalized to prevent deletion when being referenced by an active Gateway.

PR #219, opened 2021-02-24
Finalizes Gateway Dependent Resources

danehans:
Manages finalizers for dependent resources, i.e. Contour, of a Gateway. Ensures the finalizers are only removed when no other gateways reference the dependent resources.

Fixes https://github.com/projectcontour/contour-operator/issues/213

### DIFF
--- a/internal/objects/gateway/gateway.go
+++ b/internal/objects/gateway/gateway.go
@@ -105,3 +105,23 @@ func ContourForGateway(ctx context.Context, cli client.Client, gw *gatewayv1alph
 	}
 	return cntr, nil
 }
+
+// OtherGatewaysRefGatewayClass returns true if other gateways have the same
+// gatewayClassName as gw.
+func OtherGatewaysRefGatewayClass(ctx context.Context, cli client.Client, gw *gatewayv1alpha1.Gateway) (bool, error) {
+	gwList, err := OtherGatewaysExist(ctx, cli, gw)
+	if err != nil {
+		return false, fmt.Errorf("failed to verify if gateways other than %s/%s exist: %v", gw.Namespace, gw.Name, err)
+	}
+	if gwList != nil {
+		for _, g := range gwList.Items {
+			switch {
+			case g.Namespace == gw.Namespace && g.Name == gw.Name:
+				continue
+			case g.Spec.GatewayClassName == gw.Spec.GatewayClassName:
+				return true, nil
+			}
+		}
+	}
+	return false, nil
+}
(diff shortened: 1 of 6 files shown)


internal/objects/gateway/gateway.go
```diff
@@ -105,3 +105,25 @@ func ContourForGateway(ctx context.Context, cli client.Client, gw *gatewayv1alph
...
+
+// OtherGatewaysRefGatewayClass returns true if other gateways have the same
+// gatewayClassName as gw.
+func OtherGatewaysRefGatewayClass(ctx context.Context, cli client.Client, gw *gatewayv1alpha1.Gateway) (bool, error) {
+	gwList, err := OtherGatewaysExist(ctx, cli, gw)
+	if err != nil {
+		return false, fmt.Errorf("failed to verify if gateways other than %s/%s exist: %v", gw.Namespace, gw.Name, err)
+	}
+	if gwList != nil {
+		if len(gwList.Items) > 0 {
```
skriss, line 117: nit: no need to check the length of the slice here, if it's empty then the `for` will be a no-op

internal/objects/gatewayclass/gatewayclass.go
```diff
@@ -70,3 +70,27 @@ func ParameterRefExists(ctx context.Context, cli client.Client, name, ns string)
...
 	return nil, false, nil
 }
+
+// OtherGatewayClassesRefContour returns true if GatewayClasses other than gc reference contour.
+func OtherGatewayClassesRefContour(ctx context.Context, cli client.Client, gc *gatewayv1alpha1.GatewayClass, contour *operatorv1alpha1.Contour) (bool, error) {
+	gcList := &gatewayv1alpha1.GatewayClassList{}
+	if err := cli.List(ctx, gcList); err != nil {
+		return false, fmt.Errorf("failed to list gateways")
+	}
+	if len(gcList.Items) > 0 {
```
skriss, line 80: nit: no need to check the length of the slice here, if it's empty then the `for` will be a no-op

internal/objects/gatewayclass/gatewayclass.go
```diff
@@ -70,3 +70,27 @@ func ParameterRefExists(ctx context.Context, cli client.Client, name, ns string)
...
+		for _, g := range gcList.Items {
+			if g.Spec.ParametersRef != nil {
+				switch {
+				case g.Name == gc.Name && g.Namespace == gc.Namespace:
+					continue
+				case g.Spec.ParametersRef.Namespace == contour.Namespace &&
+					g.Spec.ParametersRef.Namespace == contour.Name &&
```
skriss, line 87: I think this should be:
```suggestion
					g.Spec.ParametersRef.Name == contour.Name &&
```

skriss (changes requested):
couple small things, still looking at the bigger controller changes

skriss (commented):
no other comments from me

danehans:
@skriss commit `ced0573` resolves your feedback.

youngnick (approved):

skriss (approved):

PR #236, opened 2021-03-02
Removes Unneeded Contour/Envoy Image Flags

danehans:
Now that the operator contains a [config pkg](https://github.com/projectcontour/contour-operator/blob/main/internal/operator/config/config.go), these commandline flags should no longer be set. This will allow a single location to manage the default Envoy image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ uninstall: manifests
 
 deploy: ## Deploy the operator to a Kubernetes cluster. This assumes a kubeconfig in ~/.kube/config
 deploy: manifests
-	cd config/manager && kustomize edit set image contour-operator=${IMAGE}:${VERSION}
+	cd config/manager
 	kustomize build config/default | kubectl apply -f -
 
 load-image: ## Load the operator image to a kind cluster
@@ -120,7 +120,7 @@ undeploy:
 
 example: ## Generate the example operator manifest.
 example:
-	cd config/manager && kustomize edit set image contour-operator=${IMAGE}:${VERSION}
+	cd config/manager
 	kustomize build config/default > examples/operator/operator.yaml
 
 test-example: ## Test the example Contour.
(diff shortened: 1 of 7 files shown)


danehans:
xref: https://kubernetes.slack.com/archives/C8XRH2R4J/p1614709965118700?thread_ts=1614708687.116300&cid=C8XRH2R4J

examples/operator/operator.yaml
```diff
@@ -6085,10 +6085,6 @@ spec:
         - --enable-leader-election
         command:
         - /contour-operator
-        - --contour-image
-        - docker.io/projectcontour/contour:main
```
skriss, line 6089: I *think* this one (`--contour-image`) is still needed, because this is what gets updated to the version-specific tag when a release is cut.

skriss: Dropping the Envoy ones is helpful, thanks!

danehans:
@skriss commit `67185b9` resolves https://github.com/projectcontour/contour-operator/pull/236#discussion_r585925005 and simplifies the workflow by removing the kustomization.yaml from the manager config.

skriss (approved):
LGTM, thanks @danehans!

danehans:
Commit `9bf4e7e` adds back `config/manager/kustomization.yaml`. Even though the file no longer performs any kustomization  of manager.yaml, it's required to build the example operator manifest using kustomize (`cd config/manager && kustomize build config/default > examples/operator/operator.yaml`)

Makefile
```diff
@@ -106,7 +106,7 @@ uninstall: manifests
 
 deploy: ## Deploy the operator to a Kubernetes cluster. This assumes a kubeconfig in ~/.kube/config
 deploy: manifests
-	cd config/manager && kustomize edit set image contour-operator=${IMAGE}:${VERSION}
+	cd config/manager
 	kustomize build config/default | kubectl apply -f -
```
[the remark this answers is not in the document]
danehans: @stevesloka could be removed with some additional work. Kustomize is still used to build the example operator manifest. I'd prefer to use Kustomize for this instead of a script. I think Kustomize may also be useful to support customizing the operator manifest for future requirements.

youngnick (approved):

Issue #214, opened 2021-02-23
Validate TLS of a Gateway


danehans:
**Please describe the problem you have**
Currently, TLS of a Gateway is not validated, but should be according to the [spec](https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gateway_types.go).

danehans:
Removing milestone until the community achieves a consensus on the Gateway API implementation plan.

PR #226, opened 2021-02-26
Adds Route Selection Validation for Gateway

danehans:
- Adds route selection validation for the Gateway API type.
- Adds unit tests for Gateway and GatewayClass validation logic.

Fixes: https://github.com/projectcontour/contour-operator/issues/215
Fixes: https://github.com/projectcontour/contour-operator/issues/234

Signed-off-by: Daneyon Hansen <daneyonhansen@gmail.com>

PR #237, opened 2021-03-03
Adds Support for Gateway Listener TLS Validation

danehans:
Adds support for validating TLS of a Gateway listener.

Fixes https://github.com/projectcontour/contour-operator/issues/214
Requires: https://github.com/projectcontour/contour-operator/pull/226

### DIFF
--- a/pkg/validation/validation_test.go
+++ b/pkg/validation/validation_test.go
@@ -11,14 +11,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package validation
+package validation_test
 
 import (
+	"context"
 	"fmt"
 	"testing"
 
 	operatorv1alpha1 "github.com/projectcontour/contour-operator/api/v1alpha1"
+	operatorclient "github.com/projectcontour/contour-operator/internal/operator/client"
 	"github.com/projectcontour/contour-operator/pkg/slice"
+	"github.com/projectcontour/contour-operator/pkg/validation"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


stevesloka:
We need to figure out who is ultimately responsible for writing status for these types of items. If the status is written back to the Gateway, then Contour should be doing that, otherwise both the operator & contour are going to compete for status info. 

I added this as a topic for the next maintainers meeting so we can discuss further.

pkg/validation/validation_test.go
```diff
@@ -11,14 +11,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package validation
+package validation_test
```
skriss, line 14: what was the thinking behind moving these tests to a separate package for this PR? I'm not necessarily opposed but I don't think we're doing this in other places and my top preference is usually to be consistent across the code base.

stevesloka: We need a single place for this logic to live since now two controllers are going to be doing the same logic on the same resources and those checks need to be identical. 

If for instance one controller says that the validation is fine, but another doesn't then they conflict and might not get written out to the user in a status. 

I'm not sure if that's moving all the code from `projectcontour/contour/internal` to `projectcontour/contour/pkg` or a separate project (i.e. `projectcontour/api`).

Note that the reasons we've kept that logic in `internal` was to allow us free will to change up the interfaces and logic (since internal isn't importable), but as soon as we publish those out we end up with more deprecation & integration work.

sunjayBhatia (commented):
seems like this is on hold, just sending a review to clear it out of my queue, please re-request if this needs it again

danehans:

Issue #228, opened 2021-03-01
Add option for Contour's `--ingress-class-name`


eplightning:
**Please describe the problem you have**

Currently there's no way to specify Contour's `--ingress-class-name` flag (https://projectcontour.io/docs/main/config/annotations/#ingress-class) when using the operator. 

This makes it impossible to deploy multiple instances of Contour which serve different subsets of ingresses (without deploying other Contour instances manually of course).

I might be missing something but I can't see any way of specifying it at this moment.

nak3:
This is a blocker of contour operator for Knative.

sunjayBhatia:
Sounds like we can maybe just add this configuration field in the top level `ContourSpec` alongside the Gateway class reference? https://github.com/projectcontour/contour-operator/blob/5aaf14eeb2a21606502406c13ceba000ff64ff86/api/v1alpha1/contour_types.go#L61-L92

If theres no objections to that placement I can put up a quick PR

stevesloka:
I think that works @sunjayBhatia. We'll need to have logic to check for duplicate ingress class names as well as duplicate gatewayclass names but can be a seperate PR (if not already there).

sunjayBhatia:
Oh actually before we do this how do you see this interfering with the `IngressClass` work as well: https://github.com/projectcontour/contour/issues/2146

Seems like this request is for the Ingress v1beta1 feature and the `--ingress-class-name` flag is already public, but since it's not exposed on the operator this is a chance to have a nicely defined API here with the differences between Ingress v1beta1 and Ingress v1

danehans:
xref: https://github.com/projectcontour/contour-operator/issues/240

danehans:
@eplightning @nak3 now that this issue is resolved, the operator should support the Knative use case. Can you please confirm and provide an update to this issue?

eplightning:
Just tested it out, I can now create multiple Contour instances with different `--ingress-class-name` flags using the Operator.

It's enough for my use case. I'm not an expert on Knative but it should be enough there as well.

Thanks

PR #239, opened 2021-03-04
Adds Contour IngressClass Name Support 

danehans:
Adds support for specifying Contour's IngressClass Name, i.e. the `--ingress-class-name` flag. If unspecified, Contour processes all ingress objects without an ingress-class annotation or with an annotation matching ingress-class=contour.

Fixes: https://github.com/projectcontour/contour-operator/issues/228


api/v1alpha1/contour_types.go
```diff
@@ -89,6 +89,17 @@ type ContourSpec struct {
 	// +kubebuilder:validation:MaxLength=253
 	// +optional
 	GatewayClassRef *string `json:"gatewayClassRef,omitempty"`
+
+	// IngressClassName is the name of the IngressClass used by Contour. If unset,
```
youngnick, line 93: I think there needs to be a bit more subtlety here, sadly. The Contour controller currently defaults to assuming it's the default in the cluster, indicated by an unset or `""` value for the `--ingress-class-name` argument/field.

I think the contract for this field needs to be:
- unset: default for cluster. The operator sets everything up so that Contour is the default (either no IngressClass created or create one and set it as the default, I can't remember which is preferred for IngressClass)
- set: the flag is set to the given value, and an IngressClass is created that fits that contract. It is *not* set as the cluster default.

The value can't default to `contour` because that will imply the second behavior when it's passed to the controller, sadly.

internal/objects/deployment/deployment.go
```diff
@@ -132,6 +134,10 @@ func DesiredDeployment(contour *operatorv1alpha1.Contour, image string) *appsv1.
 			args = append(args, fmt.Sprintf("--envoy-service-https-port=%d", port.PortNumber))
 		}
 	}
+	// Pass the IngressClass name if it's not the default.
+	if contour.Spec.IngressClassName != ContourIngressClassName {
```
youngnick, line 138: Ah, I said earlier that it couldn't have a default, if we do it like this, we can have a `contour` default, I suppose.

youngnick (changes requested):
I think we need to be really careful about this, because it can interact with the cluster default IngressClass in ways that can be surprising.

api/v1alpha1/contour_types.go
danehans: > The Contour controller currently defaults to assuming it's the default in the cluster, indicated by an unset or "" value for the --ingress-class-name argument/field.

@youngnick according to the [contour ingressclass docs](https://projectcontour.io/docs/main/config/annotations/#ingress-class), "contour" is the default value of --ingress-class-name. Therefore, I'm having the operator provide the same default behavior, using api defaulting instead of the `ingressClassName` field being a pointer. Are the contour ingressclass docs incorrect regarding "--ingress-class-name=contour" or am I misunderstanding the docs?

internal/objects/deployment/deployment.go
danehans: Yeah, I prefer being explicit with fields over using pointers.

sunjayBhatia: If someone explicitly sets the field in YAML to `""` it won't be overriden by the default in the kubebuilder validation setup right? Should we also check the value isn't `""` here?

danehans: > If someone explicitly sets the field in YAML to "" it won't be overriden by the default in the kubebuilder validation setup right? Should we also check the value isn't "" here?

@sunjayBhatia a Contour with `ingressClassName: ""` would fail api validation due to `+kubebuilder:validation:MinLength=1`.

Example:
```
$ cat <<EOF | kubectl apply -f -
> apiVersion: operator.projectcontour.io/v1alpha1
> kind: Contour
> metadata:
>   name: contour-sample
> spec:
>   ingressClassName: ""
>   networkPublishing:
>     envoy:
>       type: NodePortService
> EOF
The Contour "contour-sample" is invalid: spec.ingressClassName: Invalid value: "": spec.ingressClassName in body should be at least 1 chars long
```

sunjayBhatia: Ha! good call

api/v1alpha1/contour_types.go
stevesloka: If unset Contour processes ALL ingress objects without an ingress-class annotation or with an annotation matching `ingress-class=contour`. 

When you set the ingress-class explicitly, the Contour ONLY processes ingress objects that match that class.

stevesloka: So the default should be no ingress class, then is someone wants to configure one, they can.

danehans:
@stevesloka commit `ca392be` resolves your comments above.

api/v1alpha1/contour_types.go
youngnick: So now that this is a point, we have a clear distinction between "unset" and "set", and we can replicate the Contour behavior? If so, looks good to me.

youngnick (approved):
LGTM assuming I'm right about how everything will flow?

api/v1alpha1/contour_types.go
danehans: @youngnick yes, the latest commit makes `contour.spec.IngressClassName` optional to replicate Contour's default behavior: If unspecified, Contour processes all ingress objects without an ingress-class annotation or with an annotation matching ingress-class=contour. If specified, Contour processes ingress objects with an annotation matching the value of `contour.spec.IngressClassName`. I've updated the description of the PR accordingly.

danehans:
> LGTM assuming I'm right about how everything will flow?

@youngnick thanks for the review. Yes, I describe the details in https://github.com/projectcontour/contour-operator/pull/239#discussion_r595613359.

sunjayBhatia (approved):
LGTM, looks like it follows the outline set up in the controller repo for the flag 👍🏽

api/v1alpha1/contour_types.go
```diff
@@ -89,6 +89,20 @@ type ContourSpec struct {
...
+	// specified, Contour will only process ingress objects that match the provided
+	// class.
+	//
+	// For additional IngressClass details, refer to:
+	//   https://projectcontour.io/docs/main/config/annotations/#ingress-class
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +optional
+	IngressClassName *string `json:"ingressClassName,omitempty"`
```
stevesloka, line 105: Are there any more "ingress-y" things vs gatewayapi things we'd want to split apart into structs? We've done this with the "policy" stuff in Contour to allow ways to grow values and put then in categories. 

Just trying to think ahead before we get too far down the road.

sunjayBhatia: +1 thats a good point

danehans: @stevesloka I can't think of anything from an Ingress standpoint. Doing a quick review of the Contour ConfigMap, I don't see anything Ingress-specific that may be exposed through the Contour API. I see Contour has a `ingress-status-address` flag. Is this a commonly used flag?

As for Gateway API, Contour references a GatewayClass and vice versa. A GatewayClass is pretty minimal. Are there any other config knobs that you expect to expose from Contour to support the Gateway API?

skriss: :+1: in general that adding some nesting to the spec here will probably be helpful long-term as we add more and more toggles. I guess we still have room to move things around given that the API is alpha right now, though.

skriss (approved):
LGTM

Issue #211, opened 2021-02-23
Extend Gateway API e2e Tests


danehans:
**Please describe the problem you have**
The current gateway e2e test lacks the following for parity with Contour tests:
- Check gateway status conditions before proceeding with the test app deployment.
- Scape operator logs for errors.

danehans:
> Scrape operator logs for errors.

Requires https://github.com/projectcontour/contour-operator/issues/218 to be fixed.

PR #250, opened 2021-03-08
Improves Gateway Status

danehans:
Adds the "Admitted" status condition to the Contour API type. Previously, a gateway would  surface a "Ready=True" status condition even though the gateway was not ready to serve routes due to Contour not being provisioned. This occured because a Contour is 'Available=True" as long as the referenced GatewayClass existed and was "Admitted". The new Contour "Admitted" condition is true when the Contour references a GatewayClass that exists and "Available=True" when Contour provisioning is complete. This allows the Gateway to only surface "Ready=True" when Contour has been fully provisioned.
 

danehans:
Partially fixes https://github.com/projectcontour/contour-operator/issues/211

stevesloka:
Does this status reflect that Contour is ready to receive configuration or ready to receive requests? Without a default ingress object, listeners aren't created in Envoy and traffic can't route.

stevesloka:
This could be the `GatewayReasonListenersNotReady` status (https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gateway_types.go#L641), but if the operator sets that Contour would need to clear it and that interaction might be difficult to manage.

danehans:
> Does this status reflect that Contour is ready to receive configuration or ready to receive requests?

- The Gateway "Ready=True" condition means Contour is ready to serve traffic per the [spec](https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gateway_types.go#L613-L614).
- The pre-existing Contour "Available" condition is analogous to the Gateway "Ready" condition. Our [spec](https://github.com/projectcontour/contour-operator/blob/main/api/v1alpha1/contour_types.go#L274-L275) is more loosely defined than Gateway 'Ready". 
- The new Contour "Admitted=True" condition means the Contour references a GatewayClass and the GatewayClass exists.
- The GatewayClass "Admitted=True" condition means the GatewayClass is valid.

danehans:
> This could be the GatewayReasonListenersNotReady status (https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gateway_types.go#L641), but if the operator sets that Contour would need to clear it and that interaction might be difficult to manage.

@stevesloka I'm trying to get the base conditions (i.e. Available, Admitted, etc.) solidified before adding any more. This PR is an attempt to get the Gateway "Admitted" condition right. Previously, the implementation was not honoring the spec of this status condition since the Gateway was not truly ready to serve traffic.

stevesloka:
> The Gateway "Ready=True" condition means Contour is ready to serve traffic per the spec.

But if Envoy+Contour are running, it's not ready to serve traffic. I'm not sure how specific we need to be, but there's still no configuration applied and no listeners in Envoy created. 

That's why, to me, the status at first should be `GatewayReasonListenersNotReady` which means that there's no programming yet in Envoy to serve any traffic. Or do you see the operator setting "Ready=True" then Contour changing that later to match any errors it finds?

danehans:
xref: https://github.com/projectcontour/contour-operator/issues/254 to add gateway/gatewayclass status conditions unit tests.

internal/operator/controller/contour/controller.go
```diff
@@ -120,31 +148,6 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-					}
-				}
-				if err := status.SyncGatewayClass(ctx, r.client, gc, owned, valid); err != nil {
-					errs = append(errs, fmt.Errorf("failed to sync status for contour %s/%s: %w", req.Namespace,
-						req.Name, err))
-				}
-				if len(errs) != 0 {
-					return ctrl.Result{}, retryable.NewMaybeRetryableAggregate(errs)
-				}
-			}
```
danehans, line 147: ^ is no longer needed now that controllers are watching other objects required to calculate status.

internal/operator/controller/contour/controller.go
```diff
@@ -267,47 +271,6 @@ func (r *reconciler) ensureContour(ctx context.Context, contour *operatorv1alpha
...
-			errs = append(errs, fmt.Errorf("failed to sync status for gatewayclass %s: %w", gcRef, err))
-		} else {
-			r.log.Info("synced gatewayclass status for contour", "namespace", contour.Namespace, "name", contour.Name)
-		}
-	}
-	if err := status.SyncContour(ctx, cli, contour); err != nil {
-		errs = append(errs, fmt.Errorf("failed to sync status for contour %s/%s: %w", contour.Namespace, contour.Name, err))
-	} else {
-		r.log.Info("synced status for contour", "namespace", contour.Namespace, "name", contour.Name)
-	}
```
danehans, line 310: Same as https://github.com/projectcontour/contour-operator/pull/250/files#r592687632.

internal/operator/controller/gatewayclass/controller.go
```diff
@@ -72,20 +114,6 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-					req.Name, err)
-			}
-			if len(cntrs) > 0 {
-				for i, cntr := range cntrs {
-					if err := status.SyncContour(ctx, r.client, &cntrs[i]); err != nil {
-						return ctrl.Result{}, fmt.Errorf("failed to sync status for contour %s/%s: %w", cntr.Namespace, cntr.Name, err)
-					}
-					r.log.Info("synced contour for gatewayclass", "name", gc.Name)
-				}
-			}
```
danehans, line 88: Same as https://github.com/projectcontour/contour-operator/pull/250/files#r592687632.

internal/operator/controller/gatewayclass/controller.go
```diff
@@ -98,34 +126,22 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-				req.Name, err)
-		}
-		if len(cntrs) > 0 {
-			for i, cntr := range cntrs {
-				if err := status.SyncContour(ctx, r.client, &cntrs[i]); err != nil {
-					return ctrl.Result{}, fmt.Errorf("failed to sync status for contour %s/%s: %w", cntr.Namespace, cntr.Name, err)
-				}
-				r.log.Info("synced status for contour", "namespace", cntr.Namespace, "name", cntr.Name)
-			}
-		}
```
danehans, line 128: Same as https://github.com/projectcontour/contour-operator/pull/250/files#r592687632.

danehans:
> But if Envoy+Contour are running, it's not ready to serve traffic. 

@stevesloka IMO when provisioning has successfully completed, the gateway should surface the "Ready=True" condition. The user should now be able to create an HTTPRoute and have requests route through the gateway to the backend (barring any HTTPRoute misconfiguration).

@jpeach you led the status conditions effort upstream. Does my interpretation align with your intention of the "Ready=True" gateway status condition?

stevesloka:
I guess the confusion for me is, does `ready=true` mean ready for configuration or ready for requests? In this case, you've programmed the "ready for configuration" so if that's what we're after then this makes sense. 

If the listeners end up with errors, should Contour change this to be `ready=false`? If so, then it feels like this field is "ready for requests" and not "ready for configuration".

jpeach:
> I guess the confusion for me is, does `ready=true` mean ready for configuration or ready for requests? In this case, you've programmed the "ready for configuration" so if that's what we're after then this makes sense.

Per "GatewayConditionType", ready=true means the gateway can serve traffic. Not necessarily the traffic that corresponds to the latest config, just some arbitrary traffic.

> If the listeners end up with errors, should Contour change this to be `ready=false`? If so, then it feels like this field is "ready for requests" and not "ready for configuration".

Yes, because if the listeners are broken, then the proxy can't be serving any traffic (i.e. it's assumed to not be listening).

The "scheduled" condition is closer to what I think you mean by "ready for configuration".

youngnick:
I think this problem is coming up because Contour will, by default, not open any listeners until there is some configuration for it. In the case of the Gateway APIs, a Contour is bound to a single Gateway instance, so it won't open any listeners (and thus become `Ready`) until there is some configuration, and it's been admitted and accepted by the Contour that's part of the Gateway implementation. If there is no configuration, what happens?

This is assuming we go with what the status model we discussed at the last maintainers meeting, but if we do:
- The operator (in its function as a Gateway controller) provisions the Contour controller as part of implementing the Gateway request. It sets the `Scheduled` condition, and checks to see if all referenced *Route objects are marked as accepted. They won't be yet, as the Contour controller hasn't reconciled them.
- The Contour controller that's handling this Gateway looks at the Gateway, finds all the *Route objects it cares about, and creates Envoy config for them. Assuming they are all error-free, then it updates the status of the *Route objects, noting that they have are `Ready`.
- The Gateway controller (the operator) is watching the *Routes for status updates. Once it sees all the routes it cares about in `Ready`, then it updates the `Ready` condition on the Gateway to `true`.
- If there are any errors on the *Routes, then the respective Route is updated with a `Ready: false` Condition, and the operator will update the Gateway acordingly.
- In both cases, it's the Operator's job to update all its respective Status fields correctly, based on the values read from the *Route status.

I think this approach is the cleanest we can get:
- Gateways are reconciled by the operator (as it's the Gateway controller), and it needs read access only to *Route objects.
- *Routes are reconcilied by the Contour controller, and it needs read access only to Gateway objects.
- The *Route.Status struct is used for status coordination between the Contour controller and the Operator. This is a nice, clean API boundary that we can build clear rules and test cases around, while also keeping in mind the stated goal of having broken configuration requests not be pushed out to the Envoys.

It does have at least one disadvantage:
- There will be some duplicated logic between the two repos. We should strive to put this into a shared place (probably `projectcontour/contour/pkg`?).

danehans:
>  until there is some configuration

From a Gateway API perspective, it's an HTTPRoute that triggers the configuration. For example:
Creating an [HTTPRoute](https://github.com/projectcontour/contour-operator/blob/main/examples/gateway/kuard/kuard.yaml) causes contour to do the following:
```
time="2021-03-16T03:10:41Z" level=info msg="performing delayed update" context=contourEventHandler last_update=1m35.1722689s outstanding=1
time="2021-03-16T03:10:41Z" level=info msg="handling v3 xDS resource request" connection=2 context=xds node_id=envoy-98nx4 node_version=v1.17.1 resource_names="[]" response_nonce=2 type_url=type.googleapis.com/envoy.config.listener.v3.Listener version_info=2
time="2021-03-16T03:10:41Z" level=info msg="handling v3 xDS resource request" connection=3 context=xds node_id=envoy-98nx4 node_version=v1.17.1 resource_names="[ingress_http]" response_nonce= type_url=type.googleapis.com/envoy.config.route.v3.RouteConfiguration version_info=
time="2021-03-16T03:10:41Z" level=info msg="handling v3 xDS resource request" connection=1 context=xds node_id=envoy-98nx4 node_version=v1.17.1 resource_names="[]" response_nonce=2 type_url=type.googleapis.com/envoy.config.cluster.v3.Cluster version_info=2
time="2021-03-16T03:10:41Z" level=info msg="handling v3 xDS resource request" connection=4 context=xds node_id=envoy-98nx4 node_version=v1.17.1 resource_names="[projectcontour/kuard]" response_nonce= type_url=type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment version_info=
time="2021-03-16T03:10:41Z" level=info msg="handling v3 xDS resource request" connection=3 context=xds node_id=envoy-98nx4 node_version=v1.17.1 resource_names="[ingress_http]" response_nonce=2 type_url=type.googleapis.com/envoy.config.route.v3.RouteConfiguration version_info=2
time="2021-03-16T03:10:41Z" level=info msg="handling v3 xDS resource request" connection=4 context=xds node_id=envoy-98nx4 node_version=v1.17.1 resource_names="[projectcontour/kuard]" response_nonce=15 type_url=type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment version_info=15
time="2021-03-16T03:10:45Z" level=info msg="handling v3 xDS resource request" connection=4 context=xds node_id=envoy-98nx4 node_version=v1.17.1 resource_names="[projectcontour/kuard]" response_nonce=16 type_url=type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment version_info=16
```
and what is looks like from the Envoy logs:
```
lds: add/update listener 'ingress_http'
cds: add 1 cluster(s), remove 2 cluster(s)
cds: add/update cluster 'projectcontour/kuard/80/da39a3ee5e'

danehans:
> There will be some duplicated logic between the two repos. We should strive to put this into a shared place (probably projectcontour/contour/pkg?).

Examples of this are `gateway.spec.listeners[].tls` validation implemented in https://github.com/projectcontour/contour-operator/pull/237 and `gateway.spec.listeners[].routes` validation implemented in https://github.com/projectcontour/contour-operator/pull/226.

danehans:
xref contour/operator gateway status implementation doc: https://docs.google.com/document/d/1fq_zSozkPeImD1SoBvxQ6_mXmtqfGCn_77UQgFn0mAY/edit?usp=sharing

danehans:
Removing milestone until the community achieves a consensus on the Gateway API implementation plan.

danehans:

Issue #256, opened 2021-03-12
Gateway Listener Ports are Ignored


danehans:
The operator ignores gateway listener ports when managing the Envoy service. Instead, the operator uses ports specified in the Contour custom resource. When a `Contour` specified a gatewayclass, the operator should use the associated gateway to configure envoy service ports.

danehans:
Removing milestone until the community achieves a consensus on the Gateway API implementation plan.

danehans:
Closing since Gateway API support is being migrated to contour.

PR #257, opened 2021-03-12
Refactors Service Management

danehans:
Refactors how services are managed by the operator to support different configuration sources, i.e. Contour and Gateway.

Requires: https://github.com/projectcontour/contour-operator/pull/248
Fixes: https://github.com/projectcontour/contour-operator/issues/256

### DIFF
--- a/internal/equality/equality_test.go
+++ b/internal/equality/equality_test.go
@@ -28,6 +28,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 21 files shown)


internal/equality/equality_test.go
```diff
@@ -42,6 +43,40 @@ var (
 	}
 )
 
+// New makes a Gateway object using the provided ns/name for the object's
+// ns/name.
+func newGateway(ns, name string) *gatewayv1alpha1.Gateway {
```
danehans, line 48: `newGateway` is nearly identical to `New()` in the internal/gateway pkg. I plan to move equality to a public pkg in the future, so I would like to avoid adding any internal deps.

danehans:

PR #269, opened 2021-03-21
Fix go format in test/e2e/*.go

nak3:
This patch fixes go fmt by:

```
go fmt ./test/e2e/*.go
```

Signed-off-by: Kenjiro Nakayama <nakayamakenjiro@gmail.com>

PR #272, opened 2021-03-23
Updates Gofmt to Cover e2e Tests

danehans:
Updates `fmt` Make target to cover e2e tests.

xref: https://github.com/projectcontour/contour-operator/pull/269

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,6 +151,7 @@ manifests: generate-contour-crds
 # Run go fmt against code
 fmt:
 	go fmt ./...
+	go fmt ./test/e2e/
 
 # Run go vet against code
 vet:


skriss (approved):
LGTM. Not totally clear on why test/e2e isn't being picked up by the existing command, maybe due to build tags?

sunjayBhatia:
> LGTM. Not totally clear on why test/e2e isn't being picked up by the existing command, maybe due to build tags?

Yeah I think that is probably why golangci-lint isn't picking it up, I bet https://github.com/projectcontour/contour-operator/pull/275 won't fail on the `test/e2e` change

skriss:
> > LGTM. Not totally clear on why test/e2e isn't being picked up by the existing command, maybe due to build tags?
> 
> Yeah I think that is probably why golangci-lint isn't picking it up, I bet #275 won't fail on the `test/e2e` change

I tested this locally and it looked like it works when the path is explicitly specified, 🤷‍♂️

sunjayBhatia:
> > > LGTM. Not totally clear on why test/e2e isn't being picked up by the existing command, maybe due to build tags?
> > 
> > 
> > Yeah I think that is probably why golangci-lint isn't picking it up, I bet #275 won't fail on the `test/e2e` change
> 
> I tested this locally and it looked like it works when the path is explicitly specified, 🤷‍♂️

Yeah this run didn't fail: https://github.com/projectcontour/contour-operator/pull/275/checks?check_run_id=2176583612

Had to run this locally to get legit errors: `./hack/golangci-lint.sh run --build-tags=e2e`

sunjayBhatia (approved):
LGTM but we still need to fix the code linting CI job to catch these sorts of things, this change is not sufficient to prevent unformatted code from being checked in

sunjayBhatia:
> LGTM but we still need to fix the code linting CI job to catch these sorts of things, this change is not sufficient to prevent unformatted code from being checked in

I can do a quick PR to fix that @skriss unless youre already working on it

skriss:
go for it, haven't started that

Issue #281, opened 2021-03-26
Namespace Does Not Get Deleted With Multiple Contours


danehans:
**What steps did you take and what happened:**
Create multiple instances of `Contour` in separate namespaces with `spec.namespace.removeOnDeletion=true`. The namespace should be deleted when the `Contour` is deleted, but it does not.

Issue #280, opened 2021-03-25
Contour Cluster Role Bindings Fail with Multiple Contours


danehans:
**What steps did you take and what happened:**
I created multiple `Contour` instances in different namespaces. The 2nd instance never became "Available=True". The Contour logs indicated an RBAC issue:
```
time="2021-03-25T21:28:03Z" level=info msg="args: [serve --incluster --xds-address=0.0.0.0 --xds-port=8001 --contour-cafile=/certs/ca.crt --contour-cert-file=/certs/tls.crt --contour-key-file=/certs/tls.key --config-path=/config/contour.yaml]"
time="2021-03-25T21:28:03Z" level=warning msg="error listing v1 custom resource definitions: customresourcedefinitions.apiextensions.k8s.io is forbidden: User \"system:serviceaccount:projectcontour-2:contour\" cannot list resource \"customresourcedefinitions\" in API group \"apiextensions.k8s.io\" at the cluster scope"
time="2021-03-25T21:28:03Z" level=info msg="Watching Service for Ingress status" envoy-service-name=envoy envoy-service-namespace=projectcontour-2
time="2021-03-25T21:28:03Z" level=info msg="starting informers" context=informers
time="2021-03-25T21:28:03Z" level=info msg="started event handler" context=contourEventHandler
time="2021-03-25T21:28:03Z" level=info msg="started HTTP server" address="0.0.0.0:8000" context=metricsvc
time="2021-03-25T21:28:03Z" level=info msg="started HTTP server" address="127.0.0.1:6060" context=debugsvc
time="2021-03-25T21:28:03Z" level=info msg="started leader election" configmapname=leader-elect configmapnamespace=projectcontour-2
time="2021-03-25T21:28:03Z" level=info msg="attempting to acquire leader lease projectcontour-2/leader-elect...\n" context=kubernetes location="leaderelection.go:243"
time="2021-03-25T21:28:03Z" level=info msg="awaiting leadership election" context=loadBalancerStatusWriter
time="2021-03-25T21:28:03Z" level=info msg="waiting for informer caches to sync" context=xds
E0325 21:28:03.399367       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1.Endpoints: failed to list *v1.Endpoints: endpoints is forbidden: User "system:serviceaccount:projectcontour-2:contour" cannot list resource "endpoints" in API group "" at the cluster scope
[...]

PR #282, opened 2021-03-26
Includes Contour Namespace in Cluster RBAC Resources

danehans:
`OtherContoursExistInSpecNs()`: Ensures the contour from the list is not the same as the `contour` parameter before comparing `spec.namespace.name` .  Fixes https://github.com/projectcontour/contour-operator/issues/281

Manages cluster-scoped RBAC resources on a per-Contour basis. Adds an e2e to test the multi-Contour use case. Fixes https://github.com/projectcontour/contour-operator/issues/280

### DIFF
--- a/internal/objects/contour/contour.go
+++ b/internal/objects/contour/contour.go
@@ -106,6 +106,10 @@ func OtherContoursExistInSpecNs(ctx context.Context, cli client.Client, contour 
 	}
 	if exist {
 		for _, c := range contours.Items {
+			if c.Name == contour.Name && c.Namespace == contour.Namespace {
+				// Skip the contour from the list that matches the provided contour.
+				continue
+			}
 			if c.Spec.Namespace.Name == contour.Spec.Namespace.Name {
 				return true, nil
 			}
--- a/internal/objects/rbac.go
+++ b/internal/objects/rbac.go
@@ -58,11 +58,14 @@ func EnsureRBAC(ctx context.Context, cli client.Client, contour *operatorv1alpha
 			certSvcAct = svcAct
 		}
 	}
-	cr, err := objcr.EnsureClusterRole(ctx, cli, ContourRbacName, contour)
+	// ClusterRole and ClusterRoleBinding resources are namespace-named to allow ownership
+	// from individual instances of Contour.
+	nsName := fmt.Sprintf("%s-%s", ContourRbacName, contour.Spec.Namespace.Name)
+	cr, err := objcr.EnsureClusterRole(ctx, cli, nsName, contour)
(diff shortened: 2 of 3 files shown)


internal/objects/contour/contour.go
```diff
@@ -106,7 +106,9 @@ func OtherContoursExistInSpecNs(ctx context.Context, cli client.Client, contour
 	}
 	if exist {
 		for _, c := range contours.Items {
-			if c.Spec.Namespace.Name == contour.Spec.Namespace.Name {
+			if c.Name != contour.Name &&
+				c.Namespace != contour.Namespace &&
+				c.Spec.Namespace.Name == contour.Spec.Namespace.Name {
```
skriss, line 111: should this be `if (c.Name != contour.Name || c.Namespace != contour.Namespace) && c.Spec.Namespace.Name == contour.Spec.Namespace.Name {`? If it has a different namespace *or* name than `contour`, then it's not the same one.

danehans: @skriss `OtherContoursExist()` will return a list of contours, including the one that's provided as input to `OtherContoursExistInSpecNs()`. L109-L110 ensures the Contour from the list is not the same as the input Contour. I could update `OtherContoursExist()` to filter the input Contour from the list so it only returns Contours that differ from the input Contour. Thoughts?

skriss: I think it should be an `||` instead of an `&&`, since the one from the list needs to differ either by name, *or* by namespace (or both) in order to not be the same as the input Contour.

skriss:
```go
func different(c1, c2 *operatorv1alpha1.Contour) bool {
    return c1.Name != c2.Name || c1.Namespace != c2.Namespace
}
```

danehans: @skriss thanks for catching this. I simplified the conditional logic in commit `38fd172`, PTAL when you have a moment.

internal/objects/contour/contour.go
```diff
@@ -106,6 +106,10 @@ func OtherContoursExistInSpecNs(ctx context.Context, cli client.Client, contour
 	}
 	if exist {
 		for _, c := range contours.Items {
+			if c.Name == contour.Name && c.Namespace == contour.Namespace {
+				// Skip the contour from the list that matches the provided contour.
+				continue
+			}
```
skriss, line 112: :+1: yeah inverting it like this definitely helps for clarity sometimes

skriss (approved):
LGTM!

youngnick (approved):

Issue #284, opened 2021-03-30
Contour-Operator crashes when Gateways of other classes exist in the cluster


robscott:
**What steps did you take and what happened:**
Followed the first steps in https://traefik.io/blog/getting-started-with-traefik-and-the-new-kubernetes-gateway-api/:

```
kubectl apply -k "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.2.0"
helm repo add traefik https://helm.traefik.io/traefik                          
helm install traefik --set experimental.kubernetesGateway.enabled=true traefik/traefik
```

Followed the first steps in https://projectcontour.io/guides/gateway-api/.
```
kubectl apply -f https://projectcontour.io/quickstart/operator.yaml
```

Contour Operator went into crash loop with this stack trace in logs:

```
2021-03-30T17:59:29.809Z	INFO	gateway_controller	reconciling	{"request": "default/traefik-gateway"}
E0330 17:59:29.825213       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 323 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x1604520, 0x2329d20)
	/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/runtime/runtime.go:74 +0x95
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/runtime/runtime.go:48 +0x86
panic(0x1604520, 0x2329d20)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/projectcontour/contour-operator/internal/objects/gateway.ContourForGateway(0x19aa0d8, 0xc00054a600, 0x19bbf28, 0xc00029b130, 0xc000096d00, 0x0, 0x0, 0xc8)
	/internal/objects/gateway/gateway.go:102 +0x9f
github.com/projectcontour/contour-operator/internal/operator/controller/gateway.(*reconciler).ensureGatewayDeleted(0xc0002b7740, 0x19aa0d8, 0xc00054a600, 0xc000096d00, 0x7, 0xc0003f9b40)
	/internal/operator/controller/gateway/controller.go:230 +0x85
[...]

PR #287, opened 2021-03-31
Ensures Owned Gateway Resources are Reconciled

danehans:
The Gateway API uses the [controller field](https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.GatewayClassSpec), to indicate resource ownership. This PR adds a predicate to gatewayclass/gateway controllers to ensure only Gateway resources that are owned by the operator are reconciled.

Fixes https://github.com/projectcontour/contour-operator/issues/284

### DIFF
--- a/internal/operator/controller/gateway/controller.go
+++ b/internal/operator/controller/gateway/controller.go
@@ -41,6 +41,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 	gatewayv1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
 )
@@ -76,10 +77,39 @@ func New(mgr manager.Manager, cfg Config) (controller.Controller, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err := c.Watch(&source.Kind{Type: &gatewayv1alpha1.Gateway{}}, &handler.EnqueueRequestForObject{}); err != nil {
+	// Only enqueue Gateway objects that reference a GatewayClass owned by the operator.
+	if err := c.Watch(&source.Kind{Type: &gatewayv1alpha1.Gateway{}}, r.enqueueRequestForOwnedGateway()); err != nil {
 		return nil, err
 	}
 	return c, nil
+}
+
+// enqueueRequestForOwnedGateway returns an event handler that maps events to
+// Gateway objects that specify reference a GatewayClass owned by the operator.
+func (r *reconciler) enqueueRequestForOwnedGateway() handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
+		gw, ok := a.(*gatewayv1alpha1.Gateway)
+		if ok {
+			ctx := context.Background()
+			gc, err := objgw.ClassForGateway(ctx, r.client, gw)
+			if err != nil {
+				return []reconcile.Request{}
+			}
+			if gc != nil {
+				// The gateway references a gatewayclass that exists and is managed
+				// by the operator, so enqueue it for reconciliation.
+				return []reconcile.Request{
+					{
+						NamespacedName: types.NamespacedName{
+							Namespace: gw.Namespace,
+							Name:      gw.Name,
+						},
+					},
+				}
+			}
+		}
+		return []reconcile.Request{}
+	})
 }
 
 func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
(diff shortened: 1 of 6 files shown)


danehans:
> @danehans FWIW I find smaller focused PRs like this a lot easier/faster to review than bigger ones that cover multiple things at once. Where possible I'd love to see us continue to break things down like this, and I think it'll result in faster review velocity because they're not so overwhelming.

@skriss thanks for the review, I'll start working on the updates you suggest. I completely agree with the PR size issue. The status refactor will be a difficult one to reduce in size since contour<>gc<>gw status are intertwined. I marked https://github.com/projectcontour/contour-operator/pull/250 as a draft since it's not a v1.14.0 release blocker.

internal/operator/controller/gatewayclass/controller.go
```diff
@@ -56,12 +57,31 @@ func New(mgr manager.Manager) (controller.Controller, error) {
...
 		return nil, err
 	}
 	return c, nil
 }
 
+// enqueueRequestForGatewayClass returns an event handler that maps events to
+// GatewayClass objects owned by the operator.
+func (r *reconciler) enqueueRequestForGatewayClass() handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
+		gc := a.(*gatewayv1alpha1.GatewayClass)
```
sunjayBhatia, line 71: nit: To be defensive, check the type assertion succeeded here

```go
gc, ok := a.(*gatewayv1alpha1.GatewayClass)
if ok && objgc.IsController(gc) {
...
}
return []reconcile.Request{}
```

Practically it should never happen, but just to be safe

internal/operator/controller/gateway/controller.go
```diff
@@ -76,12 +77,39 @@ func New(mgr manager.Manager, cfg Config) (controller.Controller, error) {
...
 		return nil, err
 	}
 	return c, nil
 }
 
+// enqueueRequestForOwnedGateway returns an event handler that maps events to
+// Gateway objects that specify reference a GatewayClass owned by the operator.
+func (r *reconciler) enqueueRequestForOwnedGateway() handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(a client.Object) []reconcile.Request {
+		gw := a.(*gatewayv1alpha1.Gateway)
```
sunjayBhatia, line 91: nit: Same defensive type assertion check here just to be safe

sunjayBhatia (commented):
Just some small nits, but I agree with @skriss' comments especially about testing

skriss:
> @skriss thanks for the review, I'll start working on the updates you suggest. I completely agree with the PR size issue. The status refactor will be a difficult one to reduce in size since contour<>gc<>gw status are intertwined. I marked #250 as a draft since it's not a v1.14.0 release blocker.

:+1: yeah and I totally get that sometimes PRs have to be bigger to make sense, but we can try to make those more of the exception than the norm.

danehans:
@skriss commit `67bbc5b` resolves your feedback. I added an e2e that creates owned/non-owned Gateway API resources and ensures that status conditions are as expected.

sunjayBhatia (commented):
e2e tests make sense to me!

danehans:
> e2e tests make sense to me, except maybe one question: should we be setting status at all on GatewayClasses and Gateways we do not own?

@sunjayBhatia the operator should not set status of GatewayClass/Gateway objects it does not own. The status of a GatewayClass  is defaulted by the api server to "Admitted=False", [xref](https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gatewayclass_types.go#L43). The status of a Gateway is defaulted by the api server to "Scheduled=False", [xref](https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gateway_types.go#L45). The e2e test added by this PR checks the non-owned Gateway API objects for these status types/values indicating status was not set by the operator.

sunjayBhatia:
> > e2e tests make sense to me, except maybe one question: should we be setting status at all on GatewayClasses and Gateways we do not own?
> 
> @sunjayBhatia the operator should not set status of GatewayClass/Gateway objects it does not own. The status of a GatewayClass is defaulted by the api server to "Admitted=False", [xref](https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gatewayclass_types.go#L43). The status of a Gateway is defaulted by the api server to "Scheduled=False", [xref](https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gateway_types.go#L45). The e2e test added by this PR checks the non-owned Gateway API objects for these status types/values indicating status was not set by the operator.

Ah perfect that’s great context, the status defaulting makes it clear

maybe the absolute strongest assertion would be to check the status and the last transition time haven’t changed but that is probably not necessary

sunjayBhatia (approved):

internal/operator/controller/gateway/controller.go
```diff
@@ -76,12 +77,41 @@ func New(mgr manager.Manager, cfg Config) (controller.Controller, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err := c.Watch(&source.Kind{Type: &gatewayv1alpha1.Gateway{}}, &handler.EnqueueRequestForObject{}); err != nil {
+	// Only enqueue Gateway objects that reference a GatewayClass owned by the operator.
```
youngnick, line 80: Nit: The New function comment should also be updated here.

youngnick (approved):
This all LGTM. I think the status testing in general will need more love and less boilerplate, but I guess that goes on my list now :).

skriss (approved):
LGTM

Issue #290, opened 2021-04-02
operator does not clean up envoy service with ClusterIPService


nak3:
**What steps did you take and what happened:**

#### 1. Deploy contour CR with ClusterIPService.

```
$ kubectl create ns foo

$ cat <<EOF | kubectl apply -f -
apiVersion: operator.projectcontour.io/v1alpha1
kind: Contour
metadata:
  name: contour-sample
  namespace: foo
spec:
  namespace:
    name: foo
  ingressClassName: contour-internal-bar
  networkPublishing:
    envoy:
      type: ClusterIPService
EOF
```

#### 2. delete contour CR.

```
$ kubectl delete contours.operator.projectcontour.io -n foo contour-sample
```

#### 3. envoy service is not cleaned up

```
$ kubectl get svc -n foo
NAME    TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)          AGE
envoy   ClusterIP   172.20.24.61   <none>        80/TCP,443/TCP   22m
```

**What did you expect to happen:**

- envoy service should be cleaned up.

**Environment:**

- operator build on 433691b0abddb15cfdd728c478b10b5db0e3db6c

PR #292, opened 2021-04-02
Clean up envoy service when ClusterIPService type

nak3:
operator does not clean up envoy service when envoy
is ClusterIPService type.

This patch fixes it.

Fixes https://github.com/projectcontour/contour-operator/issues/290

Signed-off-by: Kenjiro Nakayama <nakayamakenjiro@gmail.com>

### DIFF
--- a/internal/operator/controller/contour/controller.go
+++ b/internal/operator/controller/contour/controller.go
@@ -311,7 +311,8 @@ func (r *reconciler) ensureContourDeleted(ctx context.Context, contour *operator
 	cli := r.client
 	if !contour.GatewayClassSet() {
 		if contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.LoadBalancerServicePublishingType ||
-			contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.NodePortServicePublishingType {
+			contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.NodePortServicePublishingType ||
+			contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.ClusterIPServicePublishingType {
 			if err := objsvc.EnsureEnvoyServiceDeleted(ctx, cli, contour); err != nil {
 				errs = append(errs, fmt.Errorf("failed to delete envoy service for contour %s/%s: %w", contour.Namespace, contour.Name, err))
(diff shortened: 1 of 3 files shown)


internal/operator/controller/contour/controller.go
```diff
@@ -311,7 +311,8 @@ func (r *reconciler) ensureContourDeleted(ctx context.Context, contour *operator
 	cli := r.client
 	if !contour.GatewayClassSet() {
 		if contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.LoadBalancerServicePublishingType ||
-			contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.NodePortServicePublishingType {
+			contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.NodePortServicePublishingType ||
```
nak3, line 314: Actually this condition may not be necessary as all Envoy types are covered.  

```
                if contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.LoadBalancerServicePublishingType ||
                        contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.NodePortServicePublishingType ||
                        contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.ClusterIPServicePublishingType {
```

If I need to delete it, I will update it with this code and around L#261.

sunjayBhatia (commented):
Seems like a change that would be good to have end to end tests for

danehans (approved):
@nak3 thanks for the fix and updating the e2e's to check for service deletion.

skriss (commented):
It'd be nice to have a version of `TestContourClusterIPService` that can test this. I see that the current version of `TestContourClusterIPService` has `RemoveNs: true` which makes this harder to test -- can we either (a) change that to `RemoveNs: false` (like the LB and NodePort tests) or (b) create a second ClusterIP test with `RemoveNs: false` so we can verify there that the Envoy service is in fact explicitly deleted?

skriss (approved):
LGTM!

skriss:
@sunjayBhatia did you have anything else here?

sunjayBhatia:
> @sunjayBhatia did you have anything else here?

Nope, seems like the e2e tests have been added 👍🏽

PR #295, opened 2021-04-05
Use PodReady rather than ContainersReady in e2e

nak3:
Current test uses ContainersReady to wait for pod condition but
PodReady makes sure the condition more strictly.
ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions


sunjayBhatia (approved):
makes sense to me to be more strict to reduce potential flakes

danehans (approved):

PR #301, opened 2021-04-12
internal/operator tests: re-get Contour before updating

skriss:
Re-gets Contour before attempting to update it, to avoid having a
stale in-memory version that can't be updated. Fixes a test flake.

### DIFF
--- a/internal/operator/suite_test.go
+++ b/internal/operator/suite_test.go
@@ -233,6 +233,7 @@ var _ = Describe("Run controller", func() {
 			}, timeout, interval).Should(Equal(gc.Name))
 
 			// Remove the GatewayClass reference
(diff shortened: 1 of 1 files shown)


skriss:
See e.g. https://github.com/projectcontour/contour-operator/runs/2324525760?check_suite_focus=true#step:5:121 for an example of the failure.

sunjayBhatia (approved):

nak3 (approved):

danehans (approved):

skriss:
Hmm, looks like this didn't actually work, I am just seeing another failure on `main` even with this change in place:
https://github.com/projectcontour/contour-operator/runs/2364617799

Issue #270, opened 2021-03-23
List instead of Get Gateway CRDs


danehans:
**Please describe the problem you have**
With https://github.com/projectcontour/contour-operator/pull/261 the operator gets each CRD required for Gateway API support. To reduce API calls, list the CRDs and iterate over the items to ensure each required Gateway CRD exists.

stevesloka:
This is how Contour does this check today: https://github.com/projectcontour/contour/blob/main/internal/k8s/clients.go#L120-L131

PR #309, opened 2021-04-17
Use rest mapper instead of Get to verify Gateway CRDs

nak3:
This patch uses rest mapper instead of `client.Get()` to verify
Gateway CRDs. It reduce API calls and is same method with Contour.

Fix https://github.com/projectcontour/contour-operator/issues/270


danehans (approved):

skriss (approved):
LGTM!

PR #311, opened 2021-04-19
Bump k8s deps for 1.21.0 release

sunjayBhatia:
See also https://github.com/projectcontour/contour/pull/3581


sunjayBhatia:
Seems blocked on https://github.com/kubernetes-sigs/controller-runtime/issues/1473

youngnick:
Hmmm, I don't think we'll be able to do this for the 1.15 release then, which is unfortunate, since the controller is already using that client-go version.

sunjayBhatia:
We could pick a main branch commit of controller runtime to use temporarily since it’s supposedly fixed there, what to people think?

skriss:
> We could pick a main branch commit of controller runtime to use temporarily since it’s supposedly fixed there, what do people think? Would be good to keep the operator usable properly on k8s 1.21 so our compatibility matrix is correct

Wondering if we even need to bump the client-go version; I'd guess that things would work fine on a 1.21 cluster even without updating the deps. If we do want to pin to a main branch commit, we should probably eyeball what else is in there to make sure we're not shooting ourselves in the foot

youngnick:
Given that the Operator is still alpha, we probably can get away with trusting to hope for the client-go compatibility for now.

sunjayBhatia:
pulled https://github.com/projectcontour/contour-operator/pull/328 out so we can remove the dep bump from consideration

danehans:
@sunjayBhatia can you resolve the merge conflict when you have a moment?

sunjayBhatia:
> @sunjayBhatia can you resolve the merge conflict when you have a moment?

focusing on https://github.com/projectcontour/contour-operator/pull/328 for a bit since that is problematic with/without the client change

skriss:
Saw there's a beta controller-runtime release out now: https://github.com/kubernetes-sigs/controller-runtime/tree/v0.9.0-beta.0

I'll leave this to @danehans to decide if he wants to take that dependency for 1.15, or wait for an official release.

go.mod
```diff
@@ -4,16 +4,16 @@ go 1.15
...
-	k8s.io/apimachinery v0.20.2
-	k8s.io/client-go v0.20.2
-	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
-	sigs.k8s.io/controller-runtime v0.8.0
+	k8s.io/api v0.21.0
+	k8s.io/apiextensions-apiserver v0.21.0
+	k8s.io/apimachinery v0.21.0
+	k8s.io/client-go v0.21.0
+	k8s.io/utils v0.0.0-20210305010621-2afb4311ab10
+	sigs.k8s.io/controller-runtime v0.9.0-alpha.1.0.20210409160505-5031262be76f
```
danehans, line 16: @sunjayBhatia v0.9.x is for testing only. I think v0.8.3 would be a better choice, unless I'm missing something that's needed in 0.9.0.

danehans (changes requested):

go.mod
sunjayBhatia: This was done because of https://github.com/kubernetes-sigs/controller-runtime/issues/1473

sunjayBhatia: Compilation fails with 0.8.3

danehans: That's a good enough reason to go with v0.9.x.

danehans:
> I'll leave this to @danehans to decide if he wants to take that dependency for 1.15, or wait for an official release.

I'm fine with alpha.

danehans (approved):

skriss (approved):
LGTM but I'd probably bump it to that new beta, just 'cause.

PR #314, opened 2021-04-19
Bump contour dependency

sunjayBhatia:
This should be automatically updated when we cut a release and possibly kept to match main in main of contour-operator


skriss:
I just looked to see where this is used and seems like the only spot is https://github.com/projectcontour/contour-operator/blob/main/internal/objects/clusterrole/cluster_role.go#L66 for the API group name. TBH I'd probably opt to just hardcode it and drop the dependency altogether, WDYT?

sunjayBhatia:
> I just looked to see where this is used and seems like the only spot is https://github.com/projectcontour/contour-operator/blob/main/internal/objects/clusterrole/cluster_role.go#L66 for the API group name. TBH I'd probably opt to just hardcode it and drop the dependency altogether, WDYT?

yeah! thats probably a better idea 👍🏽

PR #315, opened 2021-04-19
Drop contour api dependency

sunjayBhatia:
- Only used in one place, easy to hard code value
- So we don't have to keep dependency synced


skriss (approved):
LGTM

stevesloka (approved):

PR #317, opened 2021-04-20
Adds Envoy NodePort Service Configuration

### LINKED ISSUE #310: Multiple Contour fails to create NodePort envoy service due to port confliction
**What steps did you take and what happened:**

Create multiple Contour CR with `NodePortService` envoy.type.

```
kind: Contour
metadata:
  name: contour-gateway-sample
  namespace: contour-operator
spec:
  gatewayClassRef: sample-gatewayclass
  networkPublishing:
    envoy:
      type: NodePortService
---
kind: Contour
metadata:
  name: contour-external
  namespace: contour-external
spec:
  gatewayClassRef: contour-external-gatewayclass
  namespace:
    name: contour-external
  ingressClassName: contour-external
  networkPublishing:
    envoy:
      type: LoadBalancerService
```

Operator fails to create Envoy svc because currernt NodePort has fixed value `30080`.

```
2021-04-18T02:40:21.247Z	ERROR	controller-runtime.manager.controller.gateway_controller	Reconciler error	{"name": "contour", "namespace": "projectcontour", "error": "failed to get ensure gateway projectcontour/contour: failed to ensure envoy service for contour contour-operator/contour-gateway-sample: failed to create service projectcontour/envoy: Service \"envoy\" is invalid: spec.ports[0].nodePort: Invalid value: 30080: provided port is already allocated"}
```

**What did you expect to happen:**

I think there are a few solutions, for example:

A. Use NodePort with random value.
B. Keep using `30080` as default but increment the port number for second envoy service.
C. Add Port Number config in Contour CR.

danehans:
Adds support for configuring an Envoy Service of type NodePort. Previously, provisioning multiple `Contours` would fail when using "NodePortService" network publishing due to port collisions. This PR adds the following:
- A `nodePorts` field to Envoy network publishing for specifying http/https nodeports.
- Continues to use 30080/30443 as the default nodeports when `nodePorts` is unspecified.
- Updates the kind script to expose additional host ports for testing a Contour with a non-default node port.
- Updates the validation package to validate nodeports of a Contour (when specified).

Fixes #310 

cc: @nak3 


pkg/validation/validation.go
```diff
@@ -73,6 +85,38 @@ func containerPorts(contour *operatorv1alpha1.Contour) error {
 	return fmt.Errorf("http and https container ports are unspecified")
 }
 
+// nodePorts validates nodeports of contour, returning an error if the nodeports
+// do not meet the API specification.
+func nodePorts(contour *operatorv1alpha1.Contour) error {
+	if len(contour.Spec.NetworkPublishing.Envoy.NodePorts) > 0 {
```
nak3, line 91:
```suggestion
	if len(contour.Spec.NetworkPublishing.Envoy.NodePorts) == 2 {
```

nak3 (commented):
LGTM! Thank you!

danehans:
@nak3 thanks for the review. Commit `939d4da` resolves your comments above.

nak3 (approved):
/lgtm

pkg/validation/validation.go
```diff
@@ -73,6 +84,42 @@ func containerPorts(contour *operatorv1alpha1.Contour) error {
...
+// nodePorts validates nodeports of contour, returning an error if the nodeports
+// do not meet the API specification.
+func nodePorts(contour *operatorv1alpha1.Contour) error {
+	ports := contour.Spec.NetworkPublishing.Envoy.NodePorts
+	if ports == nil {
+		// When unspecified, defaults to 30080 for http and 30443 for https.
+		return nil
+	}
+
+	var numsFound []int32
```
sunjayBhatia, line 96: use maps here instead? similarly for https://github.com/projectcontour/contour-operator/blob/939d4daf8b0ce882a25f55a61d04e41736b8cc92/pkg/validation/validation.go#L61-L62

api/v1alpha1/contour_types.go
```diff
@@ -188,6 +188,28 @@ type EnvoyNetworkPublishing struct {
...
+	//
+	// For additional information on the NodePort Service, see:
+	//
+	//  https://kubernetes.io/docs/concepts/services-networking/service/#nodeport
+	//
+	// Names and port numbers must be unique in the list. Two ports must be specified,
+	// one named "http" for Envoy's insecure service and one named "https" for Envoy's
+	// secure service.
+	//
+	// TODO [danehans]: Increase maxItems when https://github.com/projectcontour/contour/pull/3263
```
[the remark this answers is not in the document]
danehans: ack, I created https://github.com/projectcontour/contour-operator/issues/322 to track this TODO and will remove from comments.

api/v1alpha1/contour_types.go
```diff
@@ -280,6 +302,27 @@ const (
...
+	// The number must:
+	//
+	// 1. Not be used by another NodePort Service.
+	// 2. Be within the cluster's nodeport service range, i.e. --service-node-port-range
+	//    flag (default: 30000-32767).
+	// 3. Be a valid network port number, i.e. greater than 0 and less than 65536.
+	//
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	PortNumber int32 `json:"portNumber"`
```
skriss, line 323: Is it possible to make this optional, i.e. to allow the Kubernetes cluster to choose the port number if not explicitly specified?

api/v1alpha1/contour_types.go
```diff
@@ -188,6 +188,28 @@ type EnvoyNetworkPublishing struct {
...
+	// created. You'll be able to contact the NodePort Service, from outside the
+	// cluster, by requesting <NodeIP>:<NodePort>.
+	//
+	// For additional information on the NodePort Service, see:
+	//
+	//  https://kubernetes.io/docs/concepts/services-networking/service/#nodeport
+	//
+	// Names and port numbers must be unique in the list. Two ports must be specified,
+	// one named "http" for Envoy's insecure service and one named "https" for Envoy's
+	// secure service.
```
youngnick, line 203: It's not clear from this if the ports must always be specified, given that `omitempty` and `+optional` are also specified. This could use some clarification.

api/v1alpha1/contour_types.go
youngnick: Yes, I think that it would make the single-Contour case easier if this was defaulted to the current standard values maybe? Unless there's something we're trying to achieve by making this explicit?

youngnick (commented):
This mostly LGTM, just needs a little more explanation about the defaulting/missing value behaviors (which it looks from the tests like there are?).

pkg/validation/validation.go
danehans: @sunjayBhatia commit `37b314c` simplifies this logic. If merged, I'll use a separate PR to convert the existing logic.

api/v1alpha1/contour_types.go
danehans: @skriss @youngnick I was trying to keep the default as-is, but I'm fine having the default select the nodeports dynamically.

danehans:
@youngnick commit `40a7e8c` resolves your feedback above, PTAL.

pkg/validation/validation.go
```diff
@@ -73,6 +84,27 @@ func containerPorts(contour *operatorv1alpha1.Contour) error {
...
+// nodePorts validates nodeports of contour, returning an error if the nodeports
+// do not meet the API specification.
+func nodePorts(contour *operatorv1alpha1.Contour) error {
+	ports := contour.Spec.NetworkPublishing.Envoy.NodePorts
+	if ports == nil {
+		// When unspecified, API server will auto-assign port numbers.
+		return nil
+	}
+
+	if ports[0].Name == ports[1].Name {
```
sunjayBhatia, line 96: should this also verify that the port names are `http` and `https`?

api/v1alpha1/contour_types.go
```diff
@@ -280,6 +302,32 @@ const (
 	GCPLoadBalancerProvider   LoadBalancerProviderType = "GCP"
 )
 
+// NodePort is the schema to specify a network port for a NodePort Service.
+type NodePort struct {
+	// Name is an IANA_SVC_NAME within the NodePort Service.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
```
sunjayBhatia, line 311: To make sure the port names are `http` or `https` could we add kubebuilder validation here? or do we want to keep the logic all in the operator itself

pkg/validation/validation_test.go
```diff
@@ -146,3 +151,114 @@ func TestContainerPorts(t *testing.T) {
 		}
 	}
 }
+
+func TestNodePorts(t *testing.T) {
+	httpPort := int32(30081)
+	httpsPort := int32(30444)
+
+	testCases := []struct {
```
sunjayBhatia, line 159: possibly need a test case for invalid port names if we don't use kubebuilder validation

test/e2e/operator_test.go
```diff
@@ -195,12 +197,26 @@ func TestDefaultContour(t *testing.T) {
...
+	// Configure the nodeports to match the kind config.
+	http := int32(30080)
+	httpPort := operatorv1alpha1.NodePort{
+		Name:       "http",
+		PortNumber: &http,
+	}
+	https := int32(30443)
+	httpsPort := operatorv1alpha1.NodePort{
+		Name:       "https",
+		PortNumber: &https,
```
sunjayBhatia, line 210: nit: could use the k8s utils `pointer` package https://pkg.go.dev/k8s.io/utils@v0.0.0-20210305010621-2afb4311ab10/pointer#Int32

test/e2e/operator_test.go
```diff
@@ -769,11 +876,25 @@ func TestGatewayOwnership(t *testing.T) {
...
+	}
+
 	cfg := objcontour.Config{
 		Name:         contourName,
 		Namespace:    operatorNs,
 		SpecNs:       specNs,
 		NetworkType:  operatorv1alpha1.NodePortServicePublishingType,
+		NodePorts:    []operatorv1alpha1.NodePort{httpsPort, httpPort},
```
sunjayBhatia, line 897: bit of a shame we have to repeat this, maybe add a helper function or variable for this if it is needed everywhere

internal/objects/service/service.go
```diff
@@ -274,8 +274,23 @@ func DesiredEnvoyService(contour *operatorv1alpha1.Contour) *corev1.Service {
...
-		svc.Spec.Ports[1].NodePort = EnvoyNodePortHTTPSPort
+		if contour.Spec.NetworkPublishing.Envoy.NodePorts == nil {
+			// Port numbers will be auto-assigned by API server.
+			httpPort := operatorv1alpha1.NodePort{Name: "http"}
+			httpsPort := operatorv1alpha1.NodePort{Name: "https"}
+			contour.Spec.NetworkPublishing.Envoy.NodePorts = []operatorv1alpha1.NodePort{httpPort, httpsPort}
+		} else {
+			// Note: API server defaulting requires min/max of 2 nodeports.
+			for _, p := range contour.Spec.NetworkPublishing.Envoy.NodePorts {
+				if p.Name == "http" && p.PortNumber != nil {
```
sunjayBhatia, line 285: this assumes only ports with the correct names come through which could still not happen, re other comments

sunjayBhatia: also doesn't this assume the `http` port is first?

by this logic is that always going to be true?

https://github.com/projectcontour/contour-operator/blob/3d5f2ab0c0b8e2fa5a3014d8a418315df54978fa/internal/objects/service/service.go#L216-L238

api/v1alpha1/contour_types.go
danehans: PR updated to resolve ^

internal/objects/service/service_test.go
```diff
@@ -140,6 +140,18 @@ func TestDesiredEnvoyService(t *testing.T) {
 		NetworkType: operatorv1alpha1.NodePortServicePublishingType,
 	}
 	cntr := objcontour.New(cfg)
+	https := int32(30443)
```
sunjayBhatia, line 143: maybe change to nonstandard port numbers for extra validation the port value comes from this setting and not the default

api/v1alpha1/contour_types.go
danehans: We could use `+kubebuilder:validation:Enum` to enforce "http" and "https" if `name` was a custom type. I think we want to stick with a string so the types doesn't need to change when Contour supports more than 2 listeners.

danehans:
@sunjayBhatia commit `d15db8e` resolves your feedback above.

api/v1alpha1/helpers.go
```diff
@@ -51,3 +53,17 @@ func (c *Contour) IsFinalized() bool {
 func (c *Contour) GatewayClassSet() bool {
 	return c.Spec.GatewayClassRef != nil
 }
+
+// MakeNodePorts returns a nodeport slice using the ports key as the nodeport name
+// and the ports value as the nodeport number.
+func MakeNodePorts(ports map[string]int) []NodePort {
```
sunjayBhatia, line 59: i would maybe argue for not putting this as a public function in the API just b/c its only used in tests right now

I think it is ok to have some duplication between packages, just I saw a lot of duplication in the e2e test file and wanted to cut down on that to make things easier to read

internal/objects/service/service_test.go
```diff
@@ -138,6 +138,7 @@ func TestDesiredEnvoyService(t *testing.T) {
 		SpecNs:      "projectcontour",
 		RemoveNs:    false,
 		NetworkType: operatorv1alpha1.NodePortServicePublishingType,
+		NodePorts:   operatorv1alpha1.MakeNodePorts(map[string]int{"http": 30080, "https": 30443}),
```
sunjayBhatia, line 141: maybe use nonstandard ports in the test to make sure they are set properly? seems like the rest of the test didn't have to change which could mean the defaults were still in use in the implementation

sunjayBhatia:
looks like the operator upgrade test flaked here too

api/v1alpha1/contour_types.go
skriss: we use the `Enum` validation on string fields in Contour, does that not work? See  https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/httpproxy.go#L371 https://github.com/projectcontour/contour/blob/main/apis/projectcontour/v1/httpproxy.go#L542

sunjayBhatia (approved):

PR #320, opened 2021-04-23
Updates e2e for non-kind clusters

danehans:
Adds support for running e2e tests against standard k8s clusters. This is required for testing features that kind does not support, i.e. AWS cloud internal load-balancer.

- Adds `isKindCluster()` to check whether the cluster under test was provisioned using kind. The func checks for the presence of the "kindnet" daemonset to make the determination.
- Adds `getWorkerNodeIP()` required to construct the test url for tests that use the "NodePortService" networking publishing type (specific to k8s clusters).
- Adds `labelWorkerNodes()` since kind does not label worker nodes with the standard "node-role.kubernetes.io/worker" label.
- Updates each test to use a unique namespace. This ensures the failure of one test does not affect other tests (running multiple Contour's in the same ns is not allowed). The `deleteNamespace()` func was update to return nil if the provided ns has a "Terminating" status. This reduces the time needed to run tests since deleting a ns may take several minutes, especially for k8s clusters.
- Adds `waitForIngressLB()` used to wait for a LB service to get assigned a hostname/ip from the cloud provider (required for k8s clusters).
- Refactors the client test pod logic into a separate func, `podWaitForHTTPResponse()`.

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -45,16 +45,10 @@ var (
 	operatorNs = "contour-operator"
 	// specNs is the spec.namespace.name of a Contour.
 	specNs = "projectcontour"
-	// testURL is the url used to test e2e functionality.
-	testURL = "http://local.projectcontour.io/"
 	// expectedDeploymentConditions are the expected status conditions of a
 	// deployment.
 	expectedDeploymentConditions = []appsv1.DeploymentCondition{
 		{Type: appsv1.DeploymentAvailable, Status: corev1.ConditionTrue},
-	}
-	// expectedPodConditions are the expected status conditions of a pod.
-	expectedPodConditions = []corev1.PodCondition{
-		{Type: corev1.PodReady, Status: corev1.ConditionTrue},
 	}
 	// expectedContourConditions are the expected status conditions of a
 	// contour.
@@ -89,6 +83,8 @@ var (
 	testAppReplicas = 3
 	// opLogMsg is the string used to search operator log messages.
 	opLogMsg = "error"
+	// isKind determines if tests should be tuned to run in a kind cluster.
+	isKind = true
 )
 
 func TestMain(m *testing.M) {
@@ -97,6 +93,17 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 	kclient = cl
+
+	isKind, err = isKindCluster(ctx, kclient)
+	if err != nil {
+		os.Exit(1)
+	}
+
+	if isKind {
+		if err := labelWorkerNodes(ctx, kclient); err != nil {
+			os.Exit(1)
+		}
+	}
 
 	os.Exit(m.Run())
 }
@@ -124,11 +131,13 @@ func TestDefaultContour(t *testing.T) {
 	}
 	t.Logf("created contour %s/%s", cntr.Namespace, cntr.Name)
 
-	svcName := "envoy"
-	if err := updateLbSvcIPAndNodePorts(ctx, kclient, 1*time.Minute, specNs, svcName); err != nil {
-		t.Fatalf("failed to update service %s/%s: %v", specNs, svcName, err)
+	if isKind {
+		svcName := "envoy"
+		if err := updateLbSvcIPAndNodePorts(ctx, kclient, 1*time.Minute, cfg.SpecNs, svcName); err != nil {
+			t.Fatalf("failed to update service %s/%s: %v", cfg.SpecNs, svcName, err)
+		}
+		t.Logf("updated service %s/%s loadbalancer IP and nodeports", cfg.SpecNs, svcName)
 	}
-	t.Logf("updated service %s/%s loadbalancer IP and nodeports", specNs, svcName)
 
 	if err := waitForContourStatusConditions(ctx, kclient, 5*time.Minute, testName, operatorNs, expectedContourConditions...); err != nil {
 		t.Fatalf("failed to observe expected status conditions for contour %s/%s: %v", operatorNs, testName, err)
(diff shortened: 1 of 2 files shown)


danehans:
fixes https://github.com/projectcontour/contour-operator/issues/321

test/e2e/utils.go
```diff
@@ -693,3 +767,97 @@ func envoyReplicas(ctx context.Context, cl client.Client, ns string) (int32, err
 	}
 	return ds.Status.NumberReady, nil
 }
+
+func waitForIngressLB(ctx context.Context, cl client.Client, timeout time.Duration, ns, name string) (string, error) {
```
youngnick, line 771: A docstring here would be nice, I think, since the behavior is not super obvious.

In addition, when you're testing on GKE, this probably won't work so well because the hostname will be the load balancer IP. We might also need to allow supplying an override hostname for the tests or something.

test/e2e/operator_test.go
```diff
@@ -418,20 +452,40 @@ func TestContourSpec(t *testing.T) {
...
+	if err := newIngress(ctx, kclient, appName, cfg.SpecNs, appName, 80); err != nil {
+		t.Fatalf("failed to create ingress %s/%s: %v", cfg.SpecNs, appName, err)
 	}
-	t.Logf("created ingress %s/%s", specNs, appName)
+	t.Logf("created ingress %s/%s", cfg.SpecNs, appName)
 
-	if err := waitForHTTPResponse(testURL, 1*time.Minute); err != nil {
-		t.Fatalf("failed to receive http response for %q: %v", testURL, err)
+	// testURL is the url used to test e2e functionality.
+	testURL := "http://local.projectcontour.io/"
```
youngnick, line 466: It looks like a lot of these tests are the same or very similar? Could we dry these up a bit, either here or in a followup?

youngnick (commented):
The overall approach looks great, but I think there are a few small updates that would really help.

I think drying up the test code would be good, there's a lot of repetition there. But that could conceivably be done in a followup.

test/e2e/operator_test.go
```diff
@@ -137,30 +145,49 @@ func TestDefaultContour(t *testing.T) {
 
 	// Create a sample workload for e2e testing.
 	appName := fmt.Sprintf("%s-%s", testAppName, testName)
-	if err := newDeployment(ctx, kclient, appName, specNs, testAppImage, testAppReplicas); err != nil {
-		t.Fatalf("failed to create deployment %s/%s: %v", specNs, appName, err)
+	if err := newDeployment(ctx, kclient, appName, cfg.SpecNs, testAppImage, testAppReplicas); err != nil {
```
sunjayBhatia, line 148: If I'm reading the test right, this is still using `projectcontour` as the namespace to run the deployment etc. in b/c the `cfg.SpecNs` value hasn't been changed from `specNs`, is that intended for the `TestDefaultContour` test?

test/e2e/operator_test.go
```diff
@@ -258,25 +305,25 @@ func TestContourNodePortService(t *testing.T) {
...
+	t.Logf("observed the deletion of namespace %s", cfg.SpecNs)
 }
 
 func TestContourClusterIPService(t *testing.T) {
 	testName := "test-clusterip-contour"
 	cfg := objcontour.Config{
 		Name:        testName,
 		Namespace:   operatorNs,
-		SpecNs:      specNs,
+		SpecNs:      fmt.Sprintf("%s-clusterip", specNs),
```
sunjayBhatia, line 326: Seems like it would be worth it to have a helper to generate a new random namespace name and to DRY this out with some more structure, here or in the future

test/e2e/operator_test.go
sunjayBhatia: +1

sunjayBhatia (commented):
This PR is going to be real useful for fixing up https://github.com/projectcontour/contour-operator/pull/311 as well, as I'm seeing flakiness in CI and locally running the e2e tests

Would be good to spend some time to DRY out this test suite and maybe look at using a test framework or growing our own to make it cleaner

Overall looks generally good, but theres a merge conflict and some questions to resolve

test/e2e/operator_test.go
```diff
@@ -45,17 +44,11 @@ var (
 	operatorNs = "contour-operator"
 	// specNs is the spec.namespace.name of a Contour.
 	specNs = "projectcontour"
```
sunjayBhatia, line 46: maybe worth just removing this variable completely so we are not tempted to use it, and to have a test helper that generates a namespace name for each test

danehans:
@sunjayBhatia @youngnick thanks for the review. Today's my last day of PTO. I'll work on resolving your comments tomorrow.

test/e2e/utils.go
danehans: > A docstring here would be nice, I think, since the behavior is not super obvious.

ack

> when you're testing on GKE, this probably won't work so well because the hostname will be the load balancer IP.

@youngnick `podWaitForHTTPResponse()` uses "local.projectcontour.io" as the host header to support hostname or IP-based ingress e2e testing.

test/e2e/operator_test.go
danehans: @sunjayBhatia that's correct.

danehans:
@youngnick @sunjayBhatia this PR is required for OpenShift to support Contour/Gateway API. Since we're at the v1.15.0 release date, I captured e2e improvements in https://github.com/projectcontour/contour-operator/issues/331. PTAL at the issue and let me know if I missed anything.

sunjayBhatia (approved):

test/e2e/operator_test.go
```diff
@@ -137,30 +146,49 @@ func TestDefaultContour(t *testing.T) {
 
 	// Create a sample workload for e2e testing.
 	appName := fmt.Sprintf("%s-%s", testAppName, testName)
-	if err := newDeployment(ctx, kclient, appName, specNs, testAppImage, testAppReplicas); err != nil {
-		t.Fatalf("failed to create deployment %s/%s: %v", specNs, appName, err)
+	if err := newDeployment(ctx, kclient, appName, cfg.SpecNs, testAppImage, testAppReplicas); err != nil {
```
skriss, line 149: :+1: glad to see this being changed, I started down this path in one of my previous PR's but abandoned it

test/e2e/operator_test.go
```diff
@@ -137,30 +146,49 @@ func TestDefaultContour(t *testing.T) {
...
+		}
+		t.Logf("ingress %s/%s has load-balancer %s", cfg.SpecNs, appName, lb)
+		testURL = fmt.Sprintf("http://%s/", lb)
+
+		// Curl the ingress from a client pod.
+		cliName := "test-client"
+		if err := podWaitForHTTPResponse(ctx, kclient, cfg.SpecNs, cliName, testURL, 3*time.Minute); err != nil {
+			t.Fatalf("failed to receive http response for %q: %v", testURL, err)
+		}
+		t.Logf("received http response for %q", testURL)
```
skriss, line 190: it'd be nice to not have to repeat this whole block of code in each test, let's see if we can abstract it.

skriss (changes requested):
I see there's an issue for tracking some E2E refactoring so my one detailed comment can be lumped under that.

The one change I would request before merging this is to give the `TestOperatorUpgrade` test a unique namespace to be deployed in - that was the only one I saw that wasn't getting its own NS. Should help avoid any issues due to namespaces taking a long time to terminate.

test/e2e/operator_test.go
danehans: @skriss I added this to https://github.com/projectcontour/contour-operator/issues/331 instead of refactoring since this PR is blocking https://github.com/projectcontour/contour-operator/pull/328 and we're so close to release.

danehans:
> The one change I would request before merging this is to give the TestOperatorUpgrade test a unique namespace to be deployed in - that was the only one I saw that wasn't getting its own NS. Should help avoid any issues due to namespaces taking a long time to terminate.

@skriss commit `529a729` resolves ^

skriss (approved):
CI's passing :shipit:

Issue #325, opened 2021-04-27
e2e tests take a long time to run


sunjayBhatia:
As we add more tests, the test run time is steadily expanding

We should investigate limiting the time these tests run as they run on each PR and commit

Currently a big bottleneck is that namespace deletion and Envoy daemonset deletion takes a long time

Ideas:
- Run each test in a more isolated manner
- Provide a configuration field for the shutdown manager so we can limit how long it waits for Envoy for and can shut down the Envoy pods quicker
- etc.

PR #328, opened 2021-04-28
Test against k8s 1.21.0 in example and e2e tests

sunjayBhatia:
Pulled out of https://github.com/projectcontour/contour-operator/pull/311 to remove the complication of a dependency bump

### DIFF
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -410,7 +410,7 @@ func TestContourClusterIPService(t *testing.T) {
 // TestContourSpec tests some spec changes such as:
 // - Enable RemoveNs.
 // - Initial replicas to 4.
-// - Increase replicas to 3.
+// - Decrease replicas to 3.
 func TestContourSpec(t *testing.T) {
 	testName := "test-user-contour"
 	cfg := objcontour.Config{
--- a/test/e2e/utils.go
+++ b/test/e2e/utils.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	gatewayv1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
@@ -513,6 +514,8 @@ func newPod(ctx context.Context, cl client.Client, ns, name, image string, cmd [
 		},
 		Spec: corev1.PodSpec{
 			Containers: []corev1.Container{c},
+			// Kill the pod immediately so it exits quickly on deletion.
+			TerminationGracePeriodSeconds: pointer.Int64Ptr(0),
 		},
 	}
 	if err := cl.Create(ctx, p); err != nil {
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -39,6 +39,10 @@ jobs:
           echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
           echo "/usr/local/kubebuilder/bin" >> $GITHUB_PATH
       - name: test-example
+        env:
+          # TODO: Use official upstream image once it is available
+          # See https://github.com/projectcontour/contour/issues/3610
+          NODEIMAGE: "stevesloka/kind-node:v1.21.0"
         run: |
           make local-cluster load-image example test-example
   test-e2e-linux:
@@ -54,6 +58,10 @@ jobs:
           echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
           echo "/usr/local/kubebuilder/bin" >> $GITHUB_PATH
       - name: test-e2e
+        env:
+          # TODO: Use official upstream image once it is available
+          # See https://github.com/projectcontour/contour/issues/3610
+          NODEIMAGE: "stevesloka/kind-node:v1.21.0"
         run: |
           make local-cluster load-image test-e2e
   codespell:
--- a/hack/actions/install-kubernetes-toolchain.sh
+++ b/hack/actions/install-kubernetes-toolchain.sh
@@ -66,6 +66,9 @@ download \
 tar -C "${DESTDIR}" -xf "${DESTDIR}/kustomize.tgz" kustomize
 rm "${DESTDIR}/kustomize.tgz"
 
-curl -L https://go.kubebuilder.io/dl/${KUBEBUILDER_VERS}/${OS}/amd64 | tar -xz -C /tmp/
+download \
+    "https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${KUBEBUILDER_VERS}/kubebuilder_${KUBEBUILDER_VERS}_${OS}_amd64.tar.gz" \
+    "/tmp/kubebuilder.tgz"
+tar -C /tmp/ -xzf /tmp/kubebuilder.tgz
 mv /tmp/kubebuilder_${KUBEBUILDER_VERS}_${OS}_amd64 kubebuilder
 sudo mv kubebuilder /usr/local/
--- a/hack/actions/install-kubernetes-toolchain.sh
+++ b/hack/actions/install-kubernetes-toolchain.sh
@@ -38,12 +38,20 @@ case "$#" in
     ;;
 esac
 
+# TODO: Remove once upstream images are available (https://github.com/projectcontour/contour/issues/3610).
+# Install latest version of kind.
+go get sigs.k8s.io/kind@master
 
-download \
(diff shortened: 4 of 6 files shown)


stevesloka (approved):

Makefile
```diff
@@ -201,4 +201,4 @@ release:
 test-e2e: ## Runs e2e tests.
 .PHONY: test-e2e
 test-e2e: deploy
-	go test -mod=readonly -timeout 20m -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
+	go test -mod=readonly -timeout 40m -count 1 -v -tags e2e -run "$(TEST)" ./test/e2e
```
danehans, line 204: @sunjayBhatia are you seeing e2e taking ~40m to complete?

sunjayBhatia: yeah, with the bump to the 1.21.0 kind image, its taking quite a while to finish, timing out deleting namespaces, taking 300s per test sometimes, if you see some of the failed runs

test/e2e/operator_test.go
```diff
@@ -450,6 +450,12 @@ func TestContourSpec(t *testing.T) {
 	}
 	t.Logf("deleted contour %s/%s", operatorNs, testName)
 
+	// Ensure the envoy service is cleaned up automatically.
```
danehans, line 453: @sunjayBhatia why are you waiting for the envoy svc deletion here?

Makefile
sunjayBhatia: its a bit better on my local machine but still really flaky, it seems to be hinged on namespace deletion taking forever

test/e2e/operator_test.go
sunjayBhatia: I thought it would help with the timeout waiting for the namespace to be deleted, but its pretty much a no-op since the service deletion has nothing to do with the pod deletion

test/e2e/operator_test.go
```diff
@@ -492,13 +498,20 @@ func TestMultipleContours(t *testing.T) {
 
 	// Ensure the default contour can be deleted and clean-up.
 	for _, testName := range testNames {
+		// Verify the user-defined namespace was removed by the operator.
+		ns := fmt.Sprintf("%s-ns", testName)
```
danehans, line 502: @sunjayBhatia this is the same as https://github.com/projectcontour/contour-operator/pull/328/files?file-filters%5B%5D=.go&file-filters%5B%5D=No+extension#diff-cf4b4e5424070a666a364d1d1b04011478d888d6d3d65c943ca7783333b7a6e4R472. Can you define this once, say after L467 and use the var throughout?

test/e2e/operator_test.go
```diff
@@ -492,13 +498,20 @@ func TestMultipleContours(t *testing.T) {
...
+
 		if err := deleteContour(ctx, kclient, 3*time.Minute, testName, operatorNs); err != nil {
 			t.Fatalf("failed to delete contour %s/%s: %v", operatorNs, testName, err)
 		}
 		t.Logf("deleted contour %s/%s", operatorNs, testName)
 
-		// Verify the user-defined namespace was removed by the operator.
-		ns := fmt.Sprintf("%s-ns", testName)
+		// Ensure the envoy service is cleaned up automatically.
+		if err := waitForServiceDeletion(ctx, kclient, 3*time.Minute, ns, "envoy"); err != nil {
```
danehans, line 510: @sunjayBhatia same comment as https://github.com/projectcontour/contour-operator/pull/328/files?file-filters%5B%5D=.go&file-filters%5B%5D=No+extension#r623288923

test/e2e/utils.go
```diff
@@ -353,6 +354,21 @@ func waitForDeploymentStatusConditions(ctx context.Context, cl client.Client, ti
 	})
 }
 
+func waitForDaemonsetUpToDate(ctx context.Context, cl client.Client, timeout time.Duration, name, ns string, numberUpToDateAndAvailable int32) error {
```
danehans, line 357: s/Daemonset/DaemonSet/

Maybe drop `UpToDate` from the name and use godocs to explain.

test/e2e/operator_test.go
```diff
@@ -1017,6 +1030,11 @@ func TestOperatorUpgrade(t *testing.T) {
 	}
 	t.Logf("observed image %s for deployment %s/contour", main, cfg.SpecNs)
 
+	if err := waitForDaemonsetUpToDate(ctx, kclient, 1*time.Minute, "envoy", cfg.SpecNs, 1); err != nil {
```
danehans, line 1033: @sunjayBhatia I don't understand the need for the add'l waits when the Contour is configured for ns removal and the tests currently have a func to wait for the removal of the test ns after the contour is deleted.

test/e2e/utils.go
```diff
@@ -353,6 +354,21 @@ func waitForDeploymentStatusConditions(ctx context.Context, cl client.Client, ti
...
+	nsName := types.NamespacedName{
+		Namespace: ns,
+		Name:      name,
+	}
+	return wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
+		daemonset := &appsv1.DaemonSet{}
+		if err := cl.Get(ctx, nsName, daemonset); err != nil {
+			return false, nil
+		}
+		return daemonset.Status.UpdatedNumberScheduled == numberUpToDateAndAvailable &&
```
danehans, line 367: Is it necessary to check both or just number of available?

Makefile
sunjayBhatia: even after there are no resources left in a namespace, I see this for a minute or so:

```
apiVersion: v1
kind: Namespace
metadata:
  creationTimestamp: "2021-04-29T18:34:46Z"
  deletionTimestamp: "2021-04-29T18:35:16Z"
  labels:
    contour.operator.projectcontour.io/owning-contour-name: test-default-contour
    contour.operator.projectcontour.io/owning-contour-namespace: contour-operator
    kubernetes.io/metadata.name: projectcontour
  name: projectcontour
  resourceVersion: "1629"
  uid: dfb0e8af-d7b5-4868-8351-0df1319bff71
spec:
  finalizers:
  - kubernetes
status:
  conditions:
  - lastTransitionTime: "2021-04-29T18:35:22Z"
    message: All resources successfully discovered
    reason: ResourcesDiscovered
    status: "False"
    type: NamespaceDeletionDiscoveryFailure
  - lastTransitionTime: "2021-04-29T18:35:22Z"
    message: All legacy kube types successfully parsed
    reason: ParsedGroupVersions
    status: "False"
    type: NamespaceDeletionGroupVersionParsingFailure
  - lastTransitionTime: "2021-04-29T18:35:22Z"
    message: All content successfully deleted, may be waiting on finalization
    reason: ContentDeleted
    status: "False"
    type: NamespaceDeletionContentFailure
  - lastTransitionTime: "2021-04-29T18:35:22Z"
    message: 'Some resources are remaining: pods. has 1 resource instances'
    reason: SomeResourcesRemain
    status: "True"
    type: NamespaceContentRemaining
  - lastTransitionTime: "2021-04-29T18:35:22Z"
    message: All content-preserving finalizers finished
    reason: ContentHasNoFinalizers
    status: "False"
    type: NamespaceFinalizersRemaining
  phase: Terminating
```

sunjayBhatia: similar status updates occur with the 1.20 kind image, but namespace deletion is much faster, on the order of seconds not minutes

test/e2e/utils.go
```diff
@@ -495,18 +513,7 @@ func deleteNamespace(ctx context.Context, cl client.Client, timeout time.Duratio
...
-
-	err := wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
-		err := cl.Get(ctx, key, ns)
-		return errors.IsNotFound(err), nil
-	})
-	if err != nil {
-		return fmt.Errorf("timed out waiting for namespace %s to be deleted: %v", ns.Name, err)
-	}
-	return nil
+	return waitForSpecNsDeletion(ctx, cl, timeout, ns.Name)
```
danehans, line 516: @sunjayBhatia having to wait for the ns to delete after every test is time consuming. It's best for tests to use unique ns's and just make sure the ns delete operation does not error.

sunjayBhatia: right, ive been trying to get to the bottom of why namespace deletion is taking so long, if it is something our code is doing or just something about the new image

lets just get #320 merged first

danehans: @sunjayBhatia it takes even longer on regular k8s clusters. It takes a while due to finalizers.

test/e2e/utils.go
sunjayBhatia: my reasoning behind checking both was to make sure if the upgrade includes an envoy bump the daemonset is updated to that version and the check won't return early etc.

sunjayBhatia:
got rid of most of the extraneous changes, probably didn't do much to help the test runtime

skriss:
CI ✅

skriss (approved):

sunjayBhatia:
> CI ✅

yeah @danehans change to use different namespaces and not wait for deletion is the thing

watching it run locally all the envoy deployments etc. stack up still but the tests pass which is good, could cause flakiness if there some timing weirdness but maybe reducing shutdown manager wait times in https://github.com/projectcontour/contour-operator/issues/325 could help

Issue #326, opened 2021-04-28
Option to pass loadBalancerIP to Contour


BostjanBozic:
**Please describe the problem you have**
It would be great to have an option to pass loadBalancerIP parameter to Contour.

This would help in a workflow, where you reserve public IP on cloud provider for Contour, configure DNS and then you can pass it in Contour configuration.

Currently only `type` parameter is supported in `providerParameters`, it would be great if this can be extended to include `loadBalancerIP`.

BostjanBozic:
This use case has been resolved by https://github.com/projectcontour/contour-operator/pull/336

jberger:
As far as I can tell, #336 doesn't allow adding generic loadBalancerIP to the contour/envoy, it only allows you to do so if you're using a specific provider. I'm running bare metal using metallb and the ability to just set loadBalancerIP on the service would be much appreciated. Am I wrong, is it possible?

youngnick:
@jberger, sorry for the delay here. You're right, the loadbalancerIP support here is only for specific providers. Would you mind opening an issue with your requirement, then we can see what we can do?

PR #330, opened 2021-04-29
Optional loadBalancerIP input in Contour

BostjanBozic:
Adds support for configuring Load Balancer IP in Envoy service. Previously, `loadBalancerIP` parameter of Envoy service was dynamically configured (from provider's side). Now user can preconfigure desired IP. This PR adds the following:
* An optional `loadBalancerIP` parameter to Contour's resource (`spec.networkPublish.envoy.loadBalancer.providerParameters`)

Updates https://github.com/projectcontour/contour-operator/issues/326

### DIFF
--- a/api/v1alpha1/contour_types.go
+++ b/api/v1alpha1/contour_types.go
@@ -266,6 +266,10 @@ type ProviderLoadBalancerParameters struct {
 	// +unionDiscriminator
 	// +kubebuilder:default=AWS
 	Type LoadBalancerProviderType `json:"type,omitempty"`
+
+	// loadBalancerIP contains IP for LoadBalancer service type, optional
+  // +optional
+	LoadBalancerIP string `json:"loadBalancerIP,omitempty"`
 }
 
 // LoadBalancerProviderType is the underlying infrastructure provider for the
--- a/api/v1alpha1/contour_types.go
+++ b/api/v1alpha1/contour_types.go
@@ -268,6 +268,7 @@ type ProviderLoadBalancerParameters struct {
 	Type LoadBalancerProviderType `json:"type,omitempty"`
 
 	// loadBalancerIP contains IP for LoadBalancer service type, optional
+	// +kubebuilder:validation:Format=ipv4
 	// +optional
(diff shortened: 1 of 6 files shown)


api/v1alpha1/contour_types.go
```diff
@@ -266,6 +266,10 @@ type ProviderLoadBalancerParameters struct {
 	// +unionDiscriminator
 	// +kubebuilder:default=AWS
 	Type LoadBalancerProviderType `json:"type,omitempty"`
+
+	// loadBalancerIP contains IP for LoadBalancer service type, optional
+	// +optional
+	LoadBalancerIP string `json:"loadBalancerIP,omitempty"`
```
sunjayBhatia, line 272: would be good to add some validation that this string is in fact an ip address (here or in the validation package) so we can give feedback to users early if they have something wrong, and some min/max length validation on this string

internal/objects/service/service.go
```diff
@@ -260,6 +260,9 @@ func DesiredEnvoyService(contour *operatorv1alpha1.Contour) *corev1.Service {
 	switch epType {
 	case operatorv1alpha1.LoadBalancerServicePublishingType:
 		svc.Spec.Type = corev1.ServiceTypeLoadBalancer
+		if len(contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.LoadBalancerIP) > 0 {
```
sunjayBhatia, line 263: would be good to have a unit test case for this

sunjayBhatia (changes requested):

api/v1alpha1/contour_types.go
BostjanBozic: Added `// +kubebuilder:validation:Format=ipv4` to have strict formating

BostjanBozic:
I am sorry, I just made a mess here, I will create new PR.

internal/objects/service/service.go
sunjayBhatia: i think that should be sufficient for now, its hard to do an end to end test for this b/c of kind + LoadBalancerService

Issue #187, opened 2021-02-02
Add AWS NLB Support


danehans:
**Please describe the problem you have**
https://github.com/projectcontour/contour-operator/pull/152 adds network publishing support. Extend `ProviderLoadBalancerParameters` to expose LB types, i.e. AWS NLB. The AWS provider type should default to the Classic ELB.

xref: https://github.com/projectcontour/contour-operator/issues/49
xref: https://kubernetes.slack.com/archives/C8XRH2R4J/p1612283957080800

cc: @Miciah

danehans:
xref: https://github.com/projectcontour/contour-operator/issues/159

Issue #327, opened 2021-04-28
Pass annotations to created resource


BostjanBozic:
**Please describe the problem you have**
It would be great if there would be a possibility to pass annotations to created objects, preferably on object level.

An example of usage:
* create `Contour` custom resource
```
apiVersion: operator.projectcontour.io/v1alpha1
kind: Contour
metadata:
  name: contour
spec:
  namespace:
    name: projectcontour
  networkPublishing:
    envoy:
      type: LoadBalancerService
      loadBalancer:
        annotation:
          service.beta.kubernetes.io/azure-load-balancer-resource-group: test-group
        scope: External
        providerParameters:
          type: Azure
```
* Following service is created
```
apiVersion: v1
kind: Service
metadata:
  annotations:
    service.beta.kubernetes.io/azure-load-balancer-resource-group: test-group
  name: envoy
  namespace: projectcontour
spec:
  loadBalancerIP: <IP>
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 8080
  - name: https
    port: 443
    protocol: TCP
    targetPort: 8443
  selector:
    app: envoy
  type: LoadBalancer
```

Example use case: use existing public IP created on Azure for envoy service.

danehans:
`providerParameters` is meant as the mechanism for passing provider-specific parameters. To provide a more predictable user experience, I prefer adding fields to `providerParameters` instead of passing annotations from the API to the Envoy Service resource directly. Your example could look something like:
```
apiVersion: operator.projectcontour.io/v1alpha1
kind: Contour
metadata:
  name: contour
spec:
  namespace:
    name: projectcontour
  networkPublishing:
    envoy:
      type: LoadBalancerService
      loadBalancer:
        scope: External
        # Optional- Assign a static IP to the LB (auto assigned by cloud provider by default).
        # This address should be from resource group "my-lb-public-ips".
        # xref: https://github.com/projectcontour/contour-operator/pull/336
        address: 200.200.200.200
        providerParameters:
          type: Azure
          azure:
            # Optional- Specifies the resource group of load balancer public IPs that aren't in the same
            # resource group as the cluster infrastructure (node resource group by default).
            resourceGroup: my-lb-public-ips
```

The result is the following Envoy Service resource is created:
```
kind: Service
metadata:
  annotations:
    service.beta.kubernetes.io/azure-load-balancer-resource-group: my-lb-public-ips
  name: envoy
  namespace: projectcontour
spec:
[...]

PR #336, opened 2021-04-30
Optional loadBalancerIP input in Contour

BostjanBozic:
Adds support for configuring Load Balancer IP in Envoy service. Previously, `loadBalancerIP` parameter of Envoy service was dynamically configured (from provider's side). Now user can preconfigure desired IP. This PR adds the following:
* option to provide cloud provider specific parameters under `Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.<cloudProvider>`
* Those parameters include LoadBalancer IP configuration as well as cloud provider specific annotations
* Unit tests for new parameters
* Validation of LoadBalancer IP parameter

Updates https://github.com/projectcontour/contour-operator/issues/326
Updates https://github.com/projectcontour/contour-operator/issues/327


api/v1alpha1/contour_types.go
```diff
@@ -266,6 +266,11 @@ type ProviderLoadBalancerParameters struct {
 	// +unionDiscriminator
 	// +kubebuilder:default=AWS
 	Type LoadBalancerProviderType `json:"type,omitempty"`
+
+	// loadBalancerIP contains IP for LoadBalancer service type, optional
+	// +kubebuilder:validation:Format=ipv4
```
sunjayBhatia, line 271: should we allow ipv6 ips as well?

BostjanBozic: I was thinking about it, but I am not sure how to handle this. Problem is that of course format can only be specified one, so e.g. I can not do `Format=ipv4;ipv6`. So I was thinking, either we move everything to `validation.go`, we have separate field (e.g. `LoadBalancerIPv6`), or maybe simplest way would be to just use same spec as it is implemented into * corev1.ServiceSpec (`json:"loadBalancerIP,omitempty" protobuf:"bytes,8,opt,name=loadBalancerIP"`).

What do you think, maybe last option would make the most sense as it follows same structure as corev1?

internal/objects/service/service.go
```diff
@@ -260,6 +260,9 @@ func DesiredEnvoyService(contour *operatorv1alpha1.Contour) *corev1.Service {
 	switch epType {
 	case operatorv1alpha1.LoadBalancerServicePublishingType:
 		svc.Spec.Type = corev1.ServiceTypeLoadBalancer
+		if len(contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.LoadBalancerIP) > 0 {
```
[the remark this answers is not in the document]
BostjanBozic: Unit test added with hardcoded IP address.

api/v1alpha1/contour_types.go
BostjanBozic: Maybe we can leave it as IPv4 for now and if there is some requirement for IPv6, we can add support later?

sunjayBhatia (commented):
I think this might need a main merge/rebase to get rid of some of the extraneous changes that crept in

internal/objects/service/service_test.go
```diff
@@ -112,6 +112,14 @@ func checkServiceHasExternalTrafficPolicy(t *testing.T, svc *corev1.Service, pol
 	}
 }
 
+func checkServiceHasLoadBalancerIP(t *testing.T, svc *corev1.Service, loadBalancerIp string) {
```
sunjayBhatia: added test seems good to me, thanks

api/v1alpha1/contour_types.go
sunjayBhatia: i would be on the side of letting ipv6 addresses be specified to start (probably via adding to the validation code we have today, probably simple enough to use `net.ParseIP`), though leaving things restricted and opening it up when needed make sense

I would like to get other @projectcontour/maintainers thoughts on this as well

BostjanBozic:
> I think this might need a main merge/rebase to get rid of some of the extraneous changes that crept in

@sunjayBhatia That is the confusing part, I actually did a rebase on my branch (as it is stated in dev guide that rebase is required before merge) and those changes crawled in.
What I did was:
```
git remote add upstream https://github.com/projectcontour/contour-operator.git
git fetch upstream
git rebase upstream/main
git push
```
Am I missing something here/doing it wrong?

api/v1alpha1/contour_types.go
danehans: I would expect `LoadBalancerIP` to be a field of `LoadBalancerStrategy` or `ProviderLoadBalancerParameters`. @BostjanBozic a few questions:
- Currently `LoadBalancerStrategy` supports LB scope. Must a loadbalancer have an external scope to specify an IP?
- Do the major cloud providers support specifying a LB IP? From the Service spec: "This feature depends on whether the underlying cloud-provider supports specifying the loadBalancerIP when a load balancer is created. This field will be ignored if the cloud-provider does not support the feature." If this feature is not widely supported, it should be part of `ProviderLoadBalancerParameters`.

danehans:
@BostjanBozic it looks like you have https://github.com/projectcontour/contour-operator/pull/317 are part of this PR. Can you please clean this up so the PR is easier to review?

api/v1alpha1/contour_types.go
BostjanBozic: @danehans Cheers for feedback. I put that under `ProviderLoadBalancerParameters`, as I was thinking this is parameter of LB itself, just for naming semantics. If preferred I can move it to `LoadBalancerStrategy` as well.
Regarding questions:
* As far as I am aware, you can also use this when `externalTrafficPolicy: Local`, but you need to provide IP in correct subnet
* I did check AWS, Azure and GCP. For Azure and GCP they support explicitly configuring `loadBalancerIP` (with Azure you might have to configure annotation in case public IP resource is configured in another resource group - https://github.com/projectcontour/contour-operator/issues/327). For AWS, you have to provide annotation `service.beta.kubernetes.io/aws-load-balancer-eip-allocations`. Good thing is if cloud provider does not support setting `loadBalancerIP`, this field will be ignored.

BostjanBozic:
> @BostjanBozic it looks like you have #317 are part of this PR. Can you please clean this up so the PR is easier to review?

@danehans Yes, it seems that after rebasing my fork, this mess occurred. Should I just checkout commit that I forked from and provide changes or actually just remove changes in my latest commit (I am worried that this PR would then overwrite work that was already done)?
Sorry for those basic questions, but I never pushed from fork into repo that was actively developed :)

danehans:
@BostjanBozic give these steps a try:

1. Reset to the head of upstream main (note: commit `790b3f5` is the most recent commit to the main branch at the time of this writing):
```
git reset 790b3f5
```

2. Remove your changes:
```
git checkout .
```

3. Make sure your branch is up-to-date with upstream (note "origin" is the name of my upstream remote):
```
git pull origin main
```

4. Git status should now show PR #342 as the latest commit for your branch:
```
$ git log --oneline
790b3f5 (HEAD -> BostjanBozic-main) Adds Envoy NodePort Service Configuration (#317)
b5e3b45 Bump Gateway API dependency (#324)
<SNIP>
```

5. Apply your commits:
```
git cherry-pick 00fc58aade01ed77dd2621d43717533ef6dbedf9
git cherry-pick 50f9f9938b649cff5fd02319f06bc859480dbfb2
git cherry-pick d6310de0db05c60c74368264dbb65243159b64dc
git cherry-pick c037a9647931e6828e395b9c8393f47a9729abd9
```

4. Git status should now show PR #342 as the latest commit for your branch:
```
$ git log --oneline
790b3f5 (HEAD -> BostjanBozic-main) Adds Envoy NodePort Service Configuration (#317)
b5e3b45 Bump Gateway API dependency (#324)
<SNIP>
```
6. Push your local branch to your remote:

api/v1alpha1/contour_types.go
danehans: `ProviderLoadBalancerParameters` is meant for passing provider-specific LB config and why I'm trying to understand what cloud providers support specifying a LB IP. If the major cloud providers support the ability to specify an IP, then the field should be in `LoadBalancerStrategy`. 

> As far as I am aware, you can also use this when externalTrafficPolicy: Local, but you need to provide IP in correct subnet

I'm referring to [this scope](https://github.com/projectcontour/contour-operator/blob/release-1.15/api/v1alpha1/contour_types.go#L253-L257), which is used to specify an internal or external LB. All major cloud providers support specifying an internal load-balancer, hence why `scope` is part of `LoadBalancerStrategy` and not `ProviderLoadBalancerParameters`. I'm trying to understand how specifying `loadBalancerIP` will work with `scope: internal`. Do cloud providers allow an internal LB scope when `loadBalancerIP` is set? If so, so any limitations exist? For example, cloud provider foo allows `loadBalancerIP` with internally-scoped LB but the address must be from "10.x.x.x/8".

danehans:
@BostjanBozic xref https://github.com/projectcontour/contour-operator/issues/327#issuecomment-832088417 for my thoughts on next steps to add this feature. I'm happy to help you through the process.

api/v1alpha1/contour_types.go
BostjanBozic: It is only Azure and GCP that are currently supporting providing exact IP. In AWS's case, binding existing LB to a service requires annotating the service. So basically following is the statue:
* Azure: you can specify `loadBalancerIP` parameter, in case of Public IP resource being in another resource group, you have to annotate service with `service.beta.kubernetes.io/azure-load-balancer-resource-group: myResourceGroup`
* GCP: you can specify `loadBalancerIP` parameter, I think nothing else is required
* AWS: specifying `loadBalancerIP` parameter has no effect, service needs annotation (`service.beta.kubernetes.io/aws-load-balancer-eip-allocations: eipalloc-<xxxxxxxxxxxxxxxxx>`)

Regarding the scope, I was only able to test Azure. Using internal scope following results occur:
* `loadBalancerIP` parameter can be configured
* `loadBalancerIP` needs to be in specific range
* when creating Public IP on Azure, this will never be in range required above, hence preconfiguring `loadBalancerIP` parameter for `Internal` scope makes no sense, at least for Azure

BostjanBozic:
@danehans Sounds good, I guess I will wait until we decide on the structure and I can review the implementation. As mentioned in https://github.com/projectcontour/contour-operator/issues/327 it would make sense to provide cloudProvider type under `providerParameters`.

Discussion docs available [here](https://docs.google.com/document/d/1VAdWaqM6PfzOiWh2o5Xtu2lb2Tc3AA6Y_2VipbQbUsY/edit?usp=sharing).

BostjanBozic:
I have set things up based on [doc](https://docs.google.com/document/d/1VAdWaqM6PfzOiWh2o5Xtu2lb2Tc3AA6Y_2VipbQbUsY), just so we can take a look how this would work out.
I have added new structure, updated `service.go` and added validation test in `service_test.go`. If required, what is missing is maybe some field validation.

internal/objects/service/service.go
```diff
@@ -262,10 +278,28 @@ func DesiredEnvoyService(contour *operatorv1alpha1.Contour) *corev1.Service {
 		svc.Spec.Type = corev1.ServiceTypeLoadBalancer
 		provider := contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type
 		lbAnnotations := LbAnnotations[provider]
+		isInternal := contour.Spec.NetworkPublishing.Envoy.LoadBalancer.Scope == operatorv1alpha1.InternalLoadBalancer
+		delete(lbAnnotations, awsLbEipAnnotation)
```
BostjanBozic, line 282: `delete` statements added, since otherwise operator carries those annotations with it once `Contour` custom resource is updated or recreated.

BostjanBozic:
@projectcontour/maintainers @sunjayBhatia @danehans Once you have time, can you check the changes and let me know if that sort of implementation would be suitable or would we change the structure?

internal/objects/service/service.go
```diff
@@ -49,10 +50,23 @@ const (
...
+	// Load Balancer IP based on desired subnet in internal LoadBalancer scope
+	azureLbSubnetAnnotation = "service.beta.kubernetes.io/azure-load-balancer-internal-subnet"
+	// azureLbResourceGroupAnnotation is a Service annotation that provides capability to
+	// assign Load Balancer IP based on Public IP Azure resource that resides in different
+	// resource group as Azure Kubernetes Service cluster in external LoadBalancer scope
+	azureLbResourceGroupAnnotation = "service.beta.kubernetes.io/azure-load-balancer-resource-group"
 	// azureProviderType is the name of the Microsoft Azure provider.
 	azureProviderType = "Azure"
+	// gcpLbSubnetAnnotation is a Service annotation that provides capability to assign
+	// Load Balancer IP to specified subnet in internal LoadBalancer scope
```
danehans, line 68: s/in internal LoadBalancer scope/when load balancer scope is set to "Internal"./. Repeat for other providers.

internal/objects/service/service.go
```diff
@@ -49,10 +50,23 @@ const (
...
+	// Balancer IP based on AWS Elastic IP resource in external LoadBalancer scope
+	awsLbEipAnnotation = "service.beta.kubernetes.io/aws-load-balancer-eip-allocations"
 	// awsProviderType is the name of the Amazon Web Services provider.
 	awsProviderType = "AWS"
+	// azureLbSubnetAnnotation is a Service annotation that provides capability to assign
+	// Load Balancer IP based on desired subnet in internal LoadBalancer scope
+	azureLbSubnetAnnotation = "service.beta.kubernetes.io/azure-load-balancer-internal-subnet"
+	// azureLbResourceGroupAnnotation is a Service annotation that provides capability to
+	// assign Load Balancer IP based on Public IP Azure resource that resides in different
+	// resource group as Azure Kubernetes Service cluster in external LoadBalancer scope
```
danehans, line 63: s/in external LoadBalancer scope/when load balancer scope is set to "External"./. Repeat for other providers.

danehans (changes requested):
This review focused on the API changes. I would like to get the API changes solidified before reviewing the implementation.

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +314,40 @@ const (
 	GCPLoadBalancerProvider   LoadBalancerProviderType = "GCP"
 )
 
+type AwsLoadBalancerParameters struct {
+	// eip is the id of created Elastic IP
+	// +kubebuilder:validation:Format=int
```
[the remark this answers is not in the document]
BostjanBozic: Done for all newly added fields.

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +314,40 @@ const (
...
+	// +optional
+	Eip int `json:"eip,omitempty"`
+}
+
+type AzureLoadBalancerParameters struct {
+	// address is the desired IP of LoadBalancer
+	// +optional
+	Address string `json:"address,omitempty"`
+
+	// resourceGroup is name of resource group where desired IP resides.
```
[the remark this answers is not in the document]
BostjanBozic: Fixed

internal/objects/service/service.go
BostjanBozic: Fixed

internal/objects/service/service.go
BostjanBozic: Resolved

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +314,40 @@ const (
...
+	// eip is the id of created Elastic IP
+	// +kubebuilder:validation:Format=int
+	// +optional
+	Eip int `json:"eip,omitempty"`
+}
+
+type AzureLoadBalancerParameters struct {
+	// address is the desired IP of LoadBalancer
+	// +optional
+	Address string `json:"address,omitempty"`
```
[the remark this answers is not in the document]
BostjanBozic: xref added to all `Address` fields as well as additional description to all fields.

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +314,40 @@ const (
 	GCPLoadBalancerProvider   LoadBalancerProviderType = "GCP"
 )
 
+type AwsLoadBalancerParameters struct {
+	// eip is the id of created Elastic IP
```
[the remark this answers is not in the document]
BostjanBozic: @danehans Thank you again for talking all this time to help me with this.
* Periods added to end of all sentences.
* I added example in the comment. I adjusted this structure a bit, so users can provide multiple Elastic IPs/their allocation IDs
* Doc links added
* You are correct, this is Allocation ID of elastic IP. I also changed type from `int` to `[]string` so that users can now input multiple allocation IDs
* Name changed from `eid` to `allocationId`
* All sentence beginnings were capitalized

internal/objects/service/service.go
```diff
@@ -49,10 +50,23 @@ const (
 	// TODO [danehans]: Make proxy protocol configurable or automatically enabled. See
 	// https://github.com/projectcontour/contour-operator/issues/49 for details.
 	awsLbBackendProtoAnnotation = "service.beta.kubernetes.io/aws-load-balancer-backend-protocol"
+	// awsLbEipAnnotation is a Service annotation that provides capability to assign Load
+	// Balancer IP based on AWS Elastic IP resource in external LoadBalancer scope
+	awsLbEipAnnotation = "service.beta.kubernetes.io/aws-load-balancer-eip-allocations"
```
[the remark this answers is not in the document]
BostjanBozic: You are correct, this only works with AWS NLBs.
Can you maybe point me to example that shows only classic LB is supported? I only found that any type of LoadBalancer parameters were so far controlled via annotations (e.g. `service.beta.kubernetes.io/aws-load-balancer-backend-protocol`). Is the reason somewhere here? Or is there lack of support for NLB from Contour itself?
Do you mean that we have to adjust annotations once deploying Envoy service, like described [here](https://projectcontour.io/guides/deploy-aws-nlb/)?

api/v1alpha1/contour_types.go
jroper: I have used GCP in this way - allocate a regional static IP address, and that set that IP address to the `loadBalancerIP` in the service, and it works. We do it this way because we have all our networking created in terraform - so we let terraform create the IP addresses, and then also create the DNS records that depend on those IP addresses. Then when we run our k8s deployment, we get the IP from terraform and plug it into our k8s deploy scripts.

jroper:
@BostjanBozic Thanks for this! It's exactly what I need!

internal/objects/service/service.go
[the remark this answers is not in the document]
danehans: Contour can work with NLB but the operator does not currently support creating the Envoy service with the "service.beta.kubernetes.io/aws-load-balancer-type: nlb" annotation. When a service is created without any annotations for the aws cloud provider, a class ELB is created (current implementation). See https://github.com/projectcontour/contour-operator/issues/187 for additional background. As I previously mentioned, NLB supported will be needed before merging this PR due to this dependency.

api/v1alpha1/contour_types.go
```diff
@@ -288,6 +288,21 @@ type ProviderLoadBalancerParameters struct {
 	// +unionDiscriminator
 	// +kubebuilder:default=AWS
 	Type LoadBalancerProviderType `json:"type,omitempty"`
+
+	// AWS specific LoadBalancer parameters.
```
danehans, line 292: I suggest adding "Present only if type is AWS" so users understand that this field has a dependency. Repeat for other providers.

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +317,75 @@ const (
 	GCPLoadBalancerProvider   LoadBalancerProviderType = "GCP"
 )
 
+type AwsLoadBalancerParameters struct {
+	// AllocationID is a list of Allocation IDs of Elastic IP addresses that are
+	// to be assigned to created Network Load Balancer.
+	// Example: "eipalloc-<xxxxxxxxxxxxxxxxx>"
```
danehans, line 323: - Add a blank line between the field description and the example.
- Although the description references "Network Load Balancer", we need to be explicit that the LB type must be NLB. Since LB type is currently not supported, we need to add NLB support before this PR so we can point to the specific LB type field.

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +317,75 @@ const (
...
+	//
+	// See: https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html
+	//
+	// +optional
+	AllocationID []string `json:"allocationId,omitempty"`
+}
+
+type AzureLoadBalancerParameters struct {
+	// Address is the desired IP of LoadBalancer. If LoadBalancerScope is Internal,
+	// IP address must reside in same virtual network as AKS and must not already
```
danehans, line 333: s/IP address/address/

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +317,75 @@ const (
...
+	// Example: "eipalloc-<xxxxxxxxxxxxxxxxx>"
+	//
+	// See: https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html
+	//
+	// +optional
+	AllocationID []string `json:"allocationId,omitempty"`
+}
+
+type AzureLoadBalancerParameters struct {
+	// Address is the desired IP of LoadBalancer. If LoadBalancerScope is Internal,
```
danehans, line 332: s/Internal/"Internal"/

internal/objects/service/service.go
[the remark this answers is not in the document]
BostjanBozic: Thanks for explanation. Should I open separate PR for that or can I include changes here?
Basic idea would be to add `loadBalancerType` field to `Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.aws` and in case it is NLB set up annotations correctly.

api/v1alpha1/contour_types.go
danehans: You may also want to mention that the address must already exist, i.e. `az network public-ip create`.

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +317,75 @@ const (
...
+	// IP address must reside in same virtual network as AKS and must not already
+	// be assigned to a resource. If IP address does not reside in same subnet as
+	// AKS, subnet parameter is also required.
+	//
+	// See:
+	// 	 https://docs.microsoft.com/en-us/azure/aks/static-ip#create-a-service-using-the-static-ip-address
+	// 	 https://docs.microsoft.com/en-us/azure/aks/internal-lb#specify-an-ip-address
+	//
+	// xref: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gateway_types.go#L492-L498
```
danehans, line 341: Sorry for the confusion, we don't want to xref in the docs. I would like to use similar kubebuilder validation tags for this field:
```
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=253
```

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +317,75 @@ const (
...
+	//
+	// +optional
+	Address string `json:"address,omitempty"`
+
+	// ResourceGroup is name of resource group where the desired IP resides. Relevant
+	// only if LoadBalancerScope is External.
```
danehans, line 347: s/External/"External"/

api/v1alpha1/contour_types.go
danehans: we want to ref the field name instead of the type in godocs: s/LoadBalancerScope/scope/ (repeat for other fields).

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +317,75 @@ const (
...
+	// Omit if desired IP is created in same resource group as k8s cluster.
+	//
+	// +optional
+	ResourceGroup string `json:"resourceGroup,omitempty"`
+
+	// Subnet is name of subnet resource where the desired IP resides. Relevant
+	// only if LoadBalancerScope is Internal and desired IP does not reside in
```
danehans, line 354: s/Internal/"Internal"/

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +317,75 @@ const (
...
+	// same subnet as AKS.
+	// Omit if desired IP is in same subnet as k8s cluster.
```
danehans, line 356: Add a blank line between 355 and 356

api/v1alpha1/contour_types.go
danehans: Can you see if we can add kubebuilder validation for the fields you're introducing, i.e. min/max length, etc..

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +317,75 @@ const (
...
+	//
+	// xref: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gateway_types.go#L492-L498
+	//
+	// +optional
+	Address string `json:"address,omitempty"`
+
+	// Subnet is name of subnet resource where the desired IP resides. Relevant only
+	// if LoadBalancerScope is Internal and desired IP does not reside in same
+	// subnet as GKE cluster.
+	// Omit if desired IP is in same subnet as k8s cluster.
```
danehans, line 381: Add a blank line between 380 and 381.

api/v1alpha1/contour_types.go
```diff
@@ -302,6 +317,75 @@ const (
...
+type GcpLoadBalancerParameters struct {
+	// Address is the desired IP of LoadBalancer. If LoadBalancerScope is Internal.
+	// IP address must reside in same subnet as GKE cluster or subnet parameter has
+	// to be provided.
+	//
+	// See:
+	// 	 https://cloud.google.com/kubernetes-engine/docs/tutorials/configuring-domain-name-static-ip#use_a_service
+	// 	 https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balancing#lb_subnet
+	//
+	// xref: https://github.com/kubernetes-sigs/gateway-api/blob/master/apis/v1alpha1/gateway_types.go#L492-L498
```
danehans, line 373: Same as https://github.com/projectcontour/contour-operator/pull/336/files#r631949465.

danehans:
@BostjanBozic thanks for the updates. I provided another round of feedback for the proposed API changes, PTAL. As I mention in my comments above, https://github.com/projectcontour/contour-operator/issues/187 will need to be resolved before we can merge this PR. Are you able to submit a separate PR that fixes https://github.com/projectcontour/contour-operator/issues/187?

BostjanBozic:
> @BostjanBozic thanks for the updates. I provided another round of feedback for the proposed API changes, PTAL. As I mention in my comments above, #187 will need to be resolved before we can merge this PR. Are you able to submit a separate PR that fixes #187?

@danehans thanks again for all your support with this. Yes, I can take a look and try to sort it out (basically support both classic and NLB for AWS).
My only question is if I can fix this as a part of this PR, as it it tights itself together? If not, I can create separate PR and I can later rebase this one.

BostjanBozic:
> @BostjanBozic Thanks for this! It's exactly what I need!

@jroper are you maybe using GCP/AWS and can maybe help me test some stuff? Otherwise I will have to build things there to perform some tests.

api/v1alpha1/contour_types.go
BostjanBozic: Fixed for all providers.

api/v1alpha1/contour_types.go
BostjanBozic: Fixed

api/v1alpha1/contour_types.go
BostjanBozic: Fixed

api/v1alpha1/contour_types.go
BostjanBozic: Fixed

api/v1alpha1/contour_types.go
BostjanBozic: I added validations based on naming limits from cloud provider's side.

api/v1alpha1/contour_types.go
BostjanBozic: Fixed

api/v1alpha1/contour_types.go
BostjanBozic: Fixed

api/v1alpha1/contour_types.go
BostjanBozic: Fixed

api/v1alpha1/contour_types.go
BostjanBozic: Fixed

api/v1alpha1/contour_types.go
BostjanBozic: Fixed

danehans:
@BostjanBozic I have some free time and will push a PR that adds support for AWS LB types. I'll link the PR and cc you.

danehans:
@BostjanBozic PTAL at https://github.com/projectcontour/contour-operator/pull/366 when you have a moment.

danehans:
@BostjanBozic I added this PR to the 1.16 milestone. Hopefully, https://github.com/projectcontour/contour-operator/pull/366 merges quickly and this follows.

BostjanBozic:
@danehans thank you again for your work. I did try to implement this as part of this PR, but I guess it makes sense from traceability stance to have it separate. Once https://github.com/projectcontour/contour-operator/pull/366 merges, I will rebase this and push updated code for review.

BostjanBozic:
@danehans I changed everything based on your PR for AWS LB Type, PTAL.
Only thing that is currently bothering me is that `address` parameter is changed, this is not propagated, which means that if you want to change LB IP, you need to recreate `Contour` object. Maybe you have an idea how this can be fixed (I tested by manually editing this parameter and this is allowed)?

api/v1alpha1/contour_types.go
```diff
@@ -345,6 +376,73 @@ const (
...
+	// Address must already exist (e.g. `az network public-ip create`).
+	//
+	// See:
+	// 	 https://docs.microsoft.com/en-us/azure/aks/static-ip#create-a-service-using-the-static-ip-address
+	// 	 https://docs.microsoft.com/en-us/azure/aks/internal-lb#specify-an-ip-address
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +optional
+	Address string `json:"address,omitempty"`
```
danehans, line 394: If Address is optional, the type should be a pointer, `*string`.

api/v1alpha1/contour_types.go
```diff
@@ -345,6 +376,73 @@ const (
...
+
+	// ResourceGroup is name of resource group where the desired IP resides. Relevant
+	// only if scope is "External".
+	//
+	// Omit if desired IP is created in same resource group as AKS cluster.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:valodation:MaxLength=90
+	// +optional
+	ResourceGroup string `json:"resourceGroup,omitempty"`
```
danehans, line 404: Same as https://github.com/projectcontour/contour-operator/pull/336/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yaml#r640185722.

api/v1alpha1/contour_types.go
```diff
@@ -345,6 +376,73 @@ const (
...
+	// if scope is "Internal" and desired IP does not reside in same subnet as AKS.
+	//
+	// Omit if desired IP is in same subnet as AKS cluster.
+	//
+	// See: https://docs.microsoft.com/en-us/azure/aks/internal-lb#specify-an-ip-address
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=80
+	// +optional
+	Subnet string `json:"subnet,omitempty"`
```
danehans, line 416: Same as https://github.com/projectcontour/contour-operator/pull/336/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yaml#r640185722.

api/v1alpha1/contour_types.go
```diff
@@ -345,6 +376,73 @@ const (
...
+	// reside in same subnet as GKE cluster or subnet parameter has to be provided.
+	//
+	// See:
+	// 	 https://cloud.google.com/kubernetes-engine/docs/tutorials/configuring-domain-name-static-ip#use_a_service
+	// 	 https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balancing#lb_subnet
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +optional
+	Address string `json:"address,omitempty"`
```
danehans, line 430: Same as https://github.com/projectcontour/contour-operator/pull/336/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yaml#r640185722.

api/v1alpha1/contour_types.go
```diff
@@ -345,6 +376,73 @@ const (
...
+	// cluster.
+	//
+	// Omit if desired IP is in same subnet as GKE cluster.
+	//
+	// See: https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balancing#lb_subnet
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	// +optional
+	Subnet string `json:"subnet,omitempty"`
```
danehans, line 443: Same as https://github.com/projectcontour/contour-operator/pull/336/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yaml#r640185722.

internal/objects/service/service.go
```diff
@@ -379,3 +426,49 @@ func backendTCPNeeded(spec *operatorv1alpha1.ContourSpec) bool {
...
+
+// allocationIDNeeded returns true if "service.beta.kubernetes.io/aws-load-balancer-eip-allocations"
+// annotation is needed based on the provided spec.
+func allocationIDNeeded(spec *operatorv1alpha1.ContourSpec) bool {
+	return spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.LoadBalancerServicePublishingType &&
+		spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type == operatorv1alpha1.AWSLoadBalancerProvider &&
+		spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.AWS != nil &&
+		spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.AWS.Type == operatorv1alpha1.AWSNetworkLoadBalancer &&
+		len(spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.AWS.AllocationID) > 0
+}
```
danehans, line 438: According to `awsLBAllocationIDAnnotation` you should also check for scope = "External".

internal/equality/equality.go
```diff
@@ -222,6 +222,11 @@ func LoadBalancerServiceChanged(current, expected *corev1.Service) (*corev1.Serv
 		changed = true
 	}
 
+	if !apiequality.Semantic.DeepEqual(current.Spec.LoadBalancerIP, expected.Spec.LoadBalancerIP) {
+		updated.Spec.LoadBalancerIP = expected.Spec.LoadBalancerIP
+		changed = true
+	}
+
```
danehans, line 229: This will cause Contours that do not specify a Service IP to fail the equality comparison. I suggest either passing in a `*operatorv1alpha1.Contour` to the func so this check is only performed on Contour objects that have a static IP assigned for provider params.

danehans:
@BostjanBozic thanks again for working on this feature and for your patience to get https://github.com/projectcontour/contour-operator/pull/366 merged. I think this PR is very close. I provided a few comments that need to get resolved before I can /approve. We're cutting v1.16.0 tomorrow. I'll check this PR before cutting the release to see if it's ready to be included in the release.

youngnick:
I don't think this one will make 1.16, pushing to 1.17.

internal/equality/equality.go
BostjanBozic: I did some testing (adding, removing `address` parameter, swapping between providers etc.) and I am not getting any errors. To me it seems that in case we do not provide `address` parameter in Contour, it will consider this as empty and will not try to add it.
Maybe I am just looking at wrong place - can you maybe point me in the right direction? Also how would you add the check for `address` parameter in this case?

BostjanBozic:
@danehans Thank you for feedback, I pushed latest changes, PTAL. Only thing that I am not sure about is equality for `LoadBalancerIP` - as mentioned in review I am not running into any errors there and at least from usability this seems to work as desired.

danehans:
> as mentioned in review I am not running into any errors there and at least from usability this seems to work as desired.

@BostjanBozic you will not see any errors because it's not an error condition. Every time the operator reconciles the `Contour`, the equality (current vs desired) comparison will fail, causing the operator to update the Envoy service. This is because we do not set the this field in [desired](https://github.com/projectcontour/contour-operator/blob/release-1.15/internal/objects/service/service.go#L213-L292). You can simulate by running the operator locally, creating a Contour, ensure it becomes  "Available" and check the `resourceVersion`. Every time the operator reconciles the `COntour`, it will update the Envoy service `resourceVersion`. Simulate a periodic reconcile by restarting the operator, check the Envoy service resource version and you should a new version.

BostjanBozic:
@danehans thank you for explanation. I am a bit lost though how can I fix this, can you maybe help me with this?
I added this part since this was the only way I was able to update `loadBalancerIP`  once I change this in Contour.

internal/equality/equality.go
danehans: @BostjanBozic one option is to update the function's signature to require a *Contour and then only call your new code if the Contour contains any static IP params introduced by this PR.

For example:
```
func LoadBalancerServiceChanged(current, expected *corev1.Service, contour *operatorv1alpha1.Contour) (*corev1.Service, bool) {
   ...
   if contour.AWS|GCP|Azure provider params specify an IP then run your equality check.
}
```
You then need to update callers of `LoadBalancerServiceChanged` to include a *Contour.

danehans:
@BostjanBozic see https://github.com/projectcontour/contour-operator/pull/336/files#r640906384.

api/v1alpha1/contour_types.go
```diff
@@ -334,6 +352,19 @@ type AWSLoadBalancerParameters struct {
...
+	// If you are using Amazon EKS 1.16 or later, you can assign Elastic IP addresses
+	// to Network Load Balancer with AllocationID. The number of Allocation IDs
+	// must match the number of subnets used for load balancer.
+	//
+	// Example: "eipalloc-<xxxxxxxxxxxxxxxxx>"
+	//
+	// See: https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html
+	//
+	// +optional
+	AllocationID *[]string `json:"allocationId,omitempty"`
```
danehans, line 367: `AllocationID` should be pluralized since the value is a list.

danehans:
@BostjanBozic please ignore https://github.com/projectcontour/contour-operator/pull/336#issuecomment-849767813. This PR looks good and should be ready to merge after fixing the merge conflicts and resolving https://github.com/projectcontour/contour-operator/pull/336#discussion_r644160055.

danehans:
@BostjanBozic can you also add unit tests to cover the changes you made to the validation pkg?

pkg/validation/validation.go
```diff
@@ -112,6 +119,26 @@ func NodePorts(contour *operatorv1alpha1.Contour) error {
 	return nil
 }
 
+// loadBalancer validates LoadBalancer "address" parameter of contour, returning an
+// error if "address" does not meet API specification.
+func loadBalancerAddress(contour *operatorv1alpha1.Contour) error {
+	if contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type == operatorv1alpha1.AzureLoadBalancerProvider &&
+		contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Azure.Address != nil {
+		validationIP := net.ParseIP(*contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Azure.Address)
```
danehans, line 127: You should also do a nil check for `contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Azure`.

pkg/validation/validation.go
```diff
@@ -112,6 +119,26 @@ func NodePorts(contour *operatorv1alpha1.Contour) error {
...
+		if validationIP == nil {
+			return fmt.Errorf("wrong LoadBalancer address format, should be string with IPv4 or IPv6 format")
+		}
+	} else if contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type == operatorv1alpha1.GCPLoadBalancerProvider &&
```
danehans, line 131: You should also do a nil check for `contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.GCP`.

danehans:
@BostjanBozic a few more comments above that need to get resolved... this PR is very close. Thanks again for your efforts and I look forward to getting this merged.

BostjanBozic:
@danehans Thanks for check again, I will get this fixed on Friday and rebase everything. For unit test they are configured in ` test/e2e/operator_test.go`  right?

danehans:
> For unit test they are configured in test/e2e/operator_test.go right?

`test/e2e/operator_test.go` are the e2e tests. Unit tests are in the accompanying go file, i.e. validation_test.go for validation.go.

danehans:
xref https://github.com/projectcontour/contour-operator/issues/379 for adding e2e tests (requires testing against cloud-specific clusters).

internal/objects/service/service.go
```diff
@@ -379,3 +426,50 @@ func backendTCPNeeded(spec *operatorv1alpha1.ContourSpec) bool {
...
+func allocationIDsNeeded(spec *operatorv1alpha1.ContourSpec) bool {
+	return spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.LoadBalancerServicePublishingType &&
+		spec.NetworkPublishing.Envoy.LoadBalancer.Scope == "External" &&
+		spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type == operatorv1alpha1.AWSLoadBalancerProvider &&
+		spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.AWS != nil &&
+		spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.AWS.Type == operatorv1alpha1.AWSNetworkLoadBalancer &&
+		spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.AWS.AllocationIDs != nil
+}
+
+//resourceGroupNeeded returns true if "service.beta.kubernetes.io/azure-load-balancer-resource-group"
```
danehans, line 441: //resourceGroupNeeded < should have a space after //

internal/objects/service/service_test.go
```diff
@@ -175,21 +187,54 @@ func TestDesiredEnvoyService(t *testing.T) {
...
+	}
+	cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters = gcpParams
+	svc = DesiredEnvoyService(cntr)
+	checkServiceHasLoadBalancerAddress(t, svc, loadBalancerAddress)
 	cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.Scope = operatorv1alpha1.InternalLoadBalancer
+	cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type = operatorv1alpha1.AWSLoadBalancerProvider
 	svc = DesiredEnvoyService(cntr)
 	checkServiceHasAnnotation(t, svc, true, awsInternalLBAnnotation)
-	cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type = operatorv1alpha1.AzureLoadBalancerProvider
+	// Check Azure internal load balancer type.
```
danehans, line 218: Space between these 2 lines.

internal/objects/service/service_test.go
```diff
@@ -175,21 +187,54 @@ func TestDesiredEnvoyService(t *testing.T) {
...
 		Type: operatorv1alpha1.AWSLoadBalancerProvider,
-		AWS:  &operatorv1alpha1.AWSLoadBalancerParameters{Type: operatorv1alpha1.AWSNetworkLoadBalancer},
+		AWS:  &operatorv1alpha1.AWSLoadBalancerParameters{Type: operatorv1alpha1.AWSNetworkLoadBalancer, AllocationIDs: &allocationIDs},
 	}
 	cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters = awsParams
 	svc = DesiredEnvoyService(cntr)
 	checkServiceHasAnnotation(t, svc, true, awsLBTypeAnnotation)
+	checkServiceHasAnnotation(t, svc, true, awsLBAllocationIDsAnnotation)
 	checkServiceHasAnnotation(t, svc, false, awsLbBackendProtoAnnotation)
+	// Check Azure external load balancer type.
```
danehans, line 197: Space before this line.

internal/objects/service/service_test.go
```diff
@@ -175,21 +187,54 @@ func TestDesiredEnvoyService(t *testing.T) {
...
+	// Check GCP internal load balancer type.
+	gcpParamsInternal := operatorv1alpha1.ProviderLoadBalancerParameters{
+		Type: operatorv1alpha1.GCPLoadBalancerProvider,
+		GCP:  &operatorv1alpha1.GCPLoadBalancerParameters{Address: &loadBalancerAddress, Subnet: &subnet},
+	}
+	cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters = gcpParamsInternal
 	svc = DesiredEnvoyService(cntr)
+	checkServiceHasLoadBalancerAddress(t, svc, loadBalancerAddress)
 	checkServiceHasAnnotation(t, svc, true, gcpLBTypeAnnotation)
+	checkServiceHasAnnotation(t, svc, true, gcpLBSubnetAnnotation)
```
danehans, line 237: `checkServiceHasAnnotation` is now `checkServiceHasAnnotations` and now accepts multiple annotations, so these lines should be merged after rebasing.

internal/objects/service/service_test.go
```diff
@@ -175,21 +187,54 @@ func TestDesiredEnvoyService(t *testing.T) {
...
+	azureParamsInternal := operatorv1alpha1.ProviderLoadBalancerParameters{
+		Type:  operatorv1alpha1.AzureLoadBalancerProvider,
+		Azure: &operatorv1alpha1.AzureLoadBalancerParameters{Address: &loadBalancerAddress, Subnet: &subnet},
+	}
+	cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters = azureParamsInternal
 	svc = DesiredEnvoyService(cntr)
+	checkServiceHasLoadBalancerAddress(t, svc, loadBalancerAddress)
 	checkServiceHasAnnotation(t, svc, true, azureInternalLBAnnotation)
-	cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type = operatorv1alpha1.GCPLoadBalancerProvider
+	checkServiceHasAnnotation(t, svc, true, azureLBSubnetAnnotation)
```
danehans, line 227: `checkServiceHasAnnotation` is now `checkServiceHasAnnotations` and now accepts multiple annotations, so these lines should be merged after rebasing.

pkg/validation/validation.go
```diff
@@ -112,6 +119,28 @@ func NodePorts(contour *operatorv1alpha1.Contour) error {
 	return nil
 }
 
+// loadBalancer validates LoadBalancer "address" parameter of contour, returning an
+// error if "address" does not meet API specification.
```
danehans, line 123: s/meet API specification/meet the API specification/

danehans (changes requested):
Most of the feedback is small nits, but there are a few areas related to testing and validation that need to get resolved. PTAL and let me know if you have any questions and thank you for your efforts in supporting the project.

BostjanBozic:
@danehans Thanks for looking into it. I went through your review and updated the code, PTAL. Maybe just a comment on changes within `equality_test.go` - I had to use different provider for my test case, as AWS does not support `address` parameter, though AWS is required for annotation test to pass. Due to that I changed to GCP there and had to init the parameter block.

api/v1alpha1/contour_types.go
```diff
@@ -345,6 +376,73 @@ const (
...
+	// +optional
+	Address *string `json:"address,omitempty"`
+
+	// ResourceGroup is the resource group name where the "address" resides. Relevant
+	// only if scope is "External".
+	//
+	// Omit if desired IP is created in same resource group as AKS cluster.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:valodation:MaxLength=90
```
danehans, line 402: s/valodation/validation/

It's surprising that `controller-gen` does not surface an error with this misspelling.

pkg/validation/validation.go
```diff
@@ -112,6 +122,49 @@ func NodePorts(contour *operatorv1alpha1.Contour) error {
 	return nil
 }
 
+// loadBalancer validates LoadBalancer "address" parameter of contour, returning an
```
danehans, line 125: s/loadBalancer/LoadBalancer/

pkg/validation/validation.go
```diff
@@ -112,6 +122,49 @@ func NodePorts(contour *operatorv1alpha1.Contour) error {
...
+		validationIP := net.ParseIP(*contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.GCP.Address)
+		if validationIP == nil {
+			return fmt.Errorf("wrong LoadBalancer address format, should be string with IPv4 or IPv6 format")
+		}
+	}
+
+	return nil
+}
+
+func LoadBalancerProvider(contour *operatorv1alpha1.Contour) error {
```
danehans, line 147: Please add godocs to this public function.

pkg/validation/validation.go
```diff
@@ -112,6 +122,49 @@ func NodePorts(contour *operatorv1alpha1.Contour) error {
...
+		}
+	} else if contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type == operatorv1alpha1.GCPLoadBalancerProvider {
+		if contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.AWS != nil ||
+			contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Azure != nil {
+			return fmt.Errorf("gcp provider chosen, other providers parameters should not be specified")
+		}
+	}
+
+	return nil
+}
```
danehans, line 166: nit: This conditional will be easier to read using a switch on the provider parameters type:
```
switch contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type {
  case operatorv1alpha1.AWSLoadBalancerProvider:
...
}

File: pkg/validation/validation_test.go [hunk 42]
```diff
@@ -155,6 +156,185 @@ func TestContainerPorts(t *testing.T) {
...
+				cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.GCP.Subnet = &testString
+			}
+		} else if tc.provider == "GCP" {
+			cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type = operatorv1alpha1.GCPLoadBalancerProvider
+			if tc.additionalProvider == "AWS" {
+				cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.AWS.AllocationIDs = strings.Split(testString, "")
+			} else if tc.additionalProvider == "Azure" {
+				cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Azure.Subnet = &testString
+			}
+		}
```
danehans, line 327: nit: This conditional will be easier to read using a switch on the provider parameters type:
```
switch tc.Provider {
  case "AWS":
...
}

File: pkg/validation/validation_test.go [hunk 43]
```diff
@@ -155,6 +156,185 @@ func TestContainerPorts(t *testing.T) {
...
+			t.Fatalf("%q: expected to fail but received no error", tc.description)
+		}
+	}
+}
+
+func TestLoadBalancerProvider(t *testing.T) {
+	testCases := []struct {
+		description        string
+		provider           string
+		additionalProvider string
```
danehans, line 234: I suggest using [LoadBalancerProviderType](https://github.com/projectcontour/contour-operator/blob/release-1.16/api/v1alpha1/contour_types.go#L286-L291) as the value instead of string for these 2 fields.

danehans (changes requested):
I think this PR is ready after these comments are resolved.

internal/equality/equality_test.go
```diff
@@ -550,12 +550,39 @@ func TestLoadBalancerServiceChanged(t *testing.T) {
...
+			expect: true,
+		},
 	}
 
 	for _, tc := range testCases {
+
 		cntr.Spec.NetworkPublishing.Envoy.Type = operatorv1alpha1.LoadBalancerServicePublishingType
 		cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.Scope = operatorv1alpha1.ExternalLoadBalancer
 		cntr.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type = operatorv1alpha1.AWSLoadBalancerProvider
+		if tc.description == "if load balancer IP changed" {
```
sunjayBhatia, line 567: might be ok to leave now, but would be good to refactor this test suite a bit

we're currently taking a global contour variable and modifying it in each test case, then generating a service from that contour, would be good to not use that global variable if we are going to modify it and for this test just create a service object to check the equality against with the copy/mutate pattern which is there and nice to avoid duplication

BostjanBozic: I was thinking the same. It would be quite a change, as it would affect whole `equality_testo.go`, so  I thought we can set this up separately.

danehans (approved):

Issue #337, opened 2021-04-30
Support ParameterRef-less GatewayClass


danehans:
**Please describe the problem you have**
`gatewayclasses.spec.parametersRef` is optional and passes validation as such. However, the operator panics for this use case:

```
2021-04-30T18:06:17.101Z	INFO	gatewayclass_controller	synced status for gatewayclass	{"name": "sample-gatewayclass"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x1369396]

goroutine 417 [running]:
github.com/projectcontour/contour-operator/internal/objects/gateway.ContourForGateway(0x1a6d918, 0xc00084f740, 0x1a7fc70, 0xc0002b3860, 0xc0000b2ea0, 0x0, 0x0, 0xc00083fb80)
	/internal/objects/gateway/gateway.go:81 +0x96
github.com/projectcontour/contour-operator/internal/operator/controller/gateway.(*reconciler).ensureGateway(0xc00013cd40, 0x1a6d918, 0xc00084f740, 0xc0000b2ea0, 0xc0000b2ea0, 0x0)
	/internal/operator/controller/gateway/controller.go:201 +0x9b
github.com/projectcontour/contour-operator/internal/operator/controller/gateway.(*reconciler).Reconcile(0xc00013cd40, 0x1a6d918, 0xc00084f740, 0xc000753e20, 0xe, 0xc000753e10, 0x7, 0xc00084f740, 0xc000030000, 0x172fc60, ...)
	/internal/operator/controller/gateway/controller.go:146 +0x3b6
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc0002ce5a0, 0x1a6d870, 0xc00013ce00, 0x16f51c0, 0xc000852120)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.0-beta.0/pkg/internal/controller/controller.go:298 +0x30d
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc0002ce5a0, 0x1a6d870, 0xc00013ce00, 0xc000577500)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.0-beta.0/pkg/internal/controller/controller.go:253 +0x205
[...]

PR #339, opened 2021-04-30
GatewayClass with missing parametersRef or missing fields is validated

sunjayBhatia:
Adds unit tests to cover this case

Removes some nil checks in internal code

Updates #337

### DIFF
--- a/pkg/validation/validation.go
+++ b/pkg/validation/validation.go
@@ -109,18 +109,20 @@ func nodePorts(contour *operatorv1alpha1.Contour) error {
 	return nil
 }
 
-// GatewayClass returns true if gc is a valid GatewayClass.
+// GatewayClass returns nil if gc is a valid GatewayClass,
+// otherwise an error.
 func GatewayClass(gc *gatewayv1alpha1.GatewayClass) error {
 	return parameterRef(gc)
 }
 
-// parameterRef returns true if parametersRef of gc is valid.
+// parameterRef returns nil if parametersRef of gc is valid,
+// otherwise an error.
 func parameterRef(gc *gatewayv1alpha1.GatewayClass) error {
 	if gc.Spec.ParametersRef == nil {
-		return nil
+		return fmt.Errorf("invalid gateway class %s, missing parametersRef", gc.Name)
 	}
 	if gc.Spec.ParametersRef.Scope == nil || *gc.Spec.ParametersRef.Scope != gatewayClassNamespacedParamRef {
-		return fmt.Errorf("invalid parametersRef for gateway class %s, only namespaced-scoped referecnes are supported", gc.Name)
+		return fmt.Errorf("invalid parametersRef for gateway class %s, only namespaced-scoped references are supported", gc.Name)
 	}
 	group := gc.Spec.ParametersRef.Group
 	if group != operatorv1alpha1.GatewayClassParamsRefGroup {
@@ -129,6 +131,9 @@ func parameterRef(gc *gatewayv1alpha1.GatewayClass) error {
 	kind := gc.Spec.ParametersRef.Kind
 	if kind != operatorv1alpha1.GatewayClassParamsRefKind {
 		return fmt.Errorf("invalid kind %q", kind)
+	}
+	if gc.Spec.ParametersRef.Namespace == nil {
+		return fmt.Errorf("invalid parametersRef for gateway class %s, missing namespace", gc.Name)
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


sunjayBhatia:
@danehans i see you opened https://github.com/projectcontour/contour-operator/issues/337

The docs say `This is optional if the controller does not require any additional configuration.`

Is there any case where one can actually create a Contour without a Contour CRD instance? If so, seems like we need some processing step to fill in values and some more testing since there are pointers all over the place

danehans:
@sunjayBhatia my thinking was that if paramRef was unspecified, Contour is provisioned with default settings, i.e. Envoy LB service, etc.. Thoughts @youngnick @stevesloka @skriss?

pkg/validation/validation.go
```diff
@@ -109,18 +109,20 @@ func nodePorts(contour *operatorv1alpha1.Contour) error {
...
 	return parameterRef(gc)
 }
 
-// parameterRef returns true if parametersRef of gc is valid.
+// parameterRef returns nil if parametersRef of gc is valid,
+// otherwise an error.
 func parameterRef(gc *gatewayv1alpha1.GatewayClass) error {
 	if gc.Spec.ParametersRef == nil {
-		return nil
+		return fmt.Errorf("invalid gateway class %s, missing parametersRef", gc.Name)
```
danehans, line 122: gatewayclass should be one word, similar to daemonset, ingressclass, etc..

pkg/validation/validation.go
```diff
@@ -109,18 +109,20 @@ func nodePorts(contour *operatorv1alpha1.Contour) error {
...
 	}
 	if gc.Spec.ParametersRef.Scope == nil || *gc.Spec.ParametersRef.Scope != gatewayClassNamespacedParamRef {
-		return fmt.Errorf("invalid parametersRef for gateway class %s, only namespaced-scoped referecnes are supported", gc.Name)
+		return fmt.Errorf("invalid parametersRef for gateway class %s, only namespaced-scoped references are supported", gc.Name)
```
danehans, line 125: same as https://github.com/projectcontour/contour-operator/pull/339/files#r625960661

pkg/validation/validation.go
```diff
@@ -130,6 +132,9 @@ func parameterRef(gc *gatewayv1alpha1.GatewayClass) error {
 	if kind != operatorv1alpha1.GatewayClassParamsRefKind {
 		return fmt.Errorf("invalid kind %q", kind)
 	}
+	if gc.Spec.ParametersRef.Namespace == nil {
+		return fmt.Errorf("invalid parametersRef for gateway class %s, missing namespace", gc.Name)
```
danehans, line 136: same as https://github.com/projectcontour/contour-operator/pull/339/files#r625960661

danehans:
@sunjayBhatia I'm fine proceeding with this PR as-is, but I would like an issue created and xref'd for handling a gc with no paramRef specified. In this scenario, the gateway would be treated the same as a default Contour. For example:
```
kind: GatewayClass
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: sample-gatewayclass
spec:
  controller: projectcontour.io/contour-operator
---
kind: Gateway
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: contour
  namespace: projectcontour
spec:
  gatewayClassName: sample-gatewayclass
  ...
```

The above is the same as:
```
kind: Contour
metadata:
  name: contour-sample
spec: {}
```

sunjayBhatia:
Sounds like this issue you created? https://github.com/projectcontour/contour-operator/issues/337

This PR was mainly to move the verbose nil checks into the validation package so we don't have to repeat them all over and to fix any crashes that could arise from fields not being checked after a resource is admitted.

I think it makes sense to support the use case you describe. This change restricts it for now and should solve panics etc. that are caused by things in a gatewayclass being nil etc. but once we refactor we can enable a nil parametersRef field/default Contour etc.

I think the issue of having to do nil checks etc. all over the code vs. rejecting resources with missing fields is an interesting one. I think because we're maybe not translating k8s resources into an internal abstraction type we have to repeat a lot of code to do nil checks etc. Could be a good way to refactor the pipeline: validate resource, translate it to internal type if valid (filling in any missing fields with defaults), use to generate further config.

danehans:
CI failing due to #341.

danehans:
> This PR was mainly to move the verbose nil checks into the validation package so we don't have to repeat them all over and to fix any crashes that could arise from fields not being checked after a resource is admitted.

I think it makes sense to support the use case you describe. This change restricts it for now and should solve panics etc. that are caused by things in a gatewayclass being nil etc. but once we refactor we can enable a nil parametersRef field/default Contour etc.

+1

danehans (approved):

PR #350, opened 2021-05-04
Bumps Gateway API CRDs

danehans:
https://github.com/projectcontour/contour-operator/pull/324 bumped gateway api to v0.3.0. This PR updates the gateway api crd's to match.

### DIFF
--- a/config/crd/gateway/01-crds.yaml
+++ b/config/crd/gateway/01-crds.yaml
@@ -2,12 +2,14 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
+    controller-gen.kubebuilder.io/version: v0.5.0
   creationTimestamp: null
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
seems like we should have a CI check to make sure these are up to date with what version we are on?

Also is there a reason we have a copy of the CRDs in this repo rather than install from upstream like here? https://github.com/projectcontour/contour/blob/fec8b13de6fd925cade6ebfd9a596db2ca1bdfe2/_integration/testsuite/make-kind-cluster.sh#L79
Is it so when you use the operator you get the Gateway API CRDs automatically?

danehans:
> Is it so when you use the operator you get the Gateway API CRDs automatically?

This is something I was considering, but Gateway API for the operator is in a holding pattern until we finalize implementation plans.

skriss (approved):
LGTM, but let's file a follow-up to add a CI check for this or pull them from the gateway-apis repo

danehans:
@skriss PTAL at https://github.com/projectcontour/contour-operator/issues/351.

youngnick (approved):

PR #366, opened 2021-05-21
Adds Support for Managing AWS Load Balancer Type

danehans:
Extends and implements the Contour API for managing AWS load balancer types. Supports Classic ELB and NLB.

Fixes https://github.com/projectcontour/contour-operator/issues/187

xref: https://github.com/projectcontour/contour-operator/pull/336

cc: @BostjanBozic


youngnick (approved):

internal/objects/service/service.go
```diff
@@ -83,7 +81,7 @@ var (
 	//  https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer
 	//
 	LbAnnotations = map[operatorv1alpha1.LoadBalancerProviderType]map[string]string{
-		awsProviderType: {
+		operatorv1alpha1.AWSLoadBalancerProvider: {
 			awsLbBackendProtoAnnotation: "tcp",
```
BostjanBozic, line 85: I believe this annotation should only be included if Classic type of Load Balancer is specified (https://projectcontour.io/guides/deploy-aws-nlb/)

internal/objects/service/service.go
```diff
@@ -94,14 +92,14 @@ var (
 	//  https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer
 	//
 	InternalLBAnnotations = map[operatorv1alpha1.LoadBalancerProviderType]map[string]string{
-		awsProviderType: {
+		operatorv1alpha1.AWSLoadBalancerProvider: {
 			awsInternalLBAnnotation: "0.0.0.0/0",
```
BostjanBozic, line 96: I believe this should be changed from `0.0.0.0/0` to `"true"`. Ref: https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer

internal/objects/service/service.go
```diff
@@ -262,6 +260,14 @@ func DesiredEnvoyService(contour *operatorv1alpha1.Contour) *corev1.Service {
 		svc.Spec.Type = corev1.ServiceTypeLoadBalancer
 		provider := contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type
 		lbAnnotations := LbAnnotations[provider]
+		// Add the NLB annotation to the map if specified by AWS provider parameters.
+		if provider == operatorv1alpha1.AWSLoadBalancerProvider &&
+			contour.Spec.NetworkPublishing.Envoy.Type == operatorv1alpha1.LoadBalancerServicePublishingType &&
```
BostjanBozic, line 265: Isn't this already tested with above `switch` clause?

internal/objects/service/service.go
```diff
@@ -262,6 +260,14 @@ func DesiredEnvoyService(contour *operatorv1alpha1.Contour) *corev1.Service {
...
+			contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.Type == operatorv1alpha1.AWSLoadBalancerProvider &&
```
BostjanBozic, line 266: Isn't this already tested in `if` clause above?

internal/objects/service/service.go
```diff
@@ -262,6 +260,14 @@ func DesiredEnvoyService(contour *operatorv1alpha1.Contour) *corev1.Service {
...
+			contour.Spec.NetworkPublishing.Envoy.LoadBalancer.ProviderParameters.AWS != nil &&
```
BostjanBozic, line 267: Would this not be obsolete, since next condition already includes `!= nil` condition in it?

BostjanBozic (changes requested):
@danehans I checked this, looks good. Only thing that I believe has to change is now `aws-load-balancer-backend-protol` annotation is handled, as it should be implemented only if `Classic` load balancer is used: https://projectcontour.io/guides/deploy-aws-nlb/

internal/objects/service/service.go
danehans: AWS provider params is an optional field. Checking to make sure it's not nil and the param type is NLB.

danehans:
@BostjanBozic commit `6c4d753` resolves your feedback above, PTAL.

BostjanBozic (approved):

BostjanBozic:
@danehans LGTM

sunjayBhatia (approved):

PR #372, opened 2021-05-27
internal/objects/service: refactor test

skriss:
Refactors to check each Envoy service for an exact set
of annotation keys, rather than checking for the existence
and non-existence of a subset of annotations.


sunjayBhatia (approved):

danehans (approved):

danehans:
Thanks for the PR @skriss.

Issue #362, opened 2021-05-19
define affinity, tolerations for envoy pods


r0bj:
**Please describe the problem you have**
[A clear, concise, description of the problem you are facing. What is the _problem_ that feature X would solve for you?]

I would like to deploy envoy to subset of my nodes with specific taints. It would be useful to have ability to set affinity, tolerations for envoy pods (maybe also for contour pods).

danehans:
@r0bj thanks for the creating issue. I like the idea of adding placement support for envoy and contour pods. Is this a feature that you can help develop? Do you mind updating this issue to be focused on this issue and create separate issues for the other requested features?

danehans:
> I would like to deploy envoy to subset of my nodes with specific taints. It would be useful to have ability to set affinity, tolerations for envoy pods (maybe also for contour pods).

xref https://github.com/projectcontour/contour-operator/pull/373 for envoy pod placement.

r0bj:
I updated this issue to cover only placement support.

r0bj:
It could be probably combined with supporting pod template: https://github.com/projectcontour/contour-operator/issues/378

danehans:
> It could be probably combined with supporting pod template: #378

I prefer that issues are as specific as possible, so let's not combine.

@r0bj I believe https://github.com/projectcontour/contour-operator/pull/373 resolves this issue. If so, please confirm and close this issue. If not, please let us know what else is needed.

r0bj:
@danehans Thanks!

PR #373, opened 2021-05-28
support running envoy on specific nodes

amdonov:
Trying to gauge willingness to accept this feature. PR is not ready to merge as is. Needs more comments and tests. We like to run envoy pods on a dedicated node pool. This would allow use of node selectors and tolerations to target specific nodes.


api/v1alpha1/contour_types.go
```diff
@@ -229,6 +229,41 @@ type EnvoyNetworkPublishing struct {
...
 	ContainerPorts []ContainerPort `json:"containerPorts,omitempty"`
+
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// +optional
+	Tolerations []Toleration `json:"tolerations,omitempty"`
+}
+
+type Toleration struct {
```
skriss, line 240: any particular reason to use a custom type here, vs. reusing the upstream `corev1.Toleration`?

api/v1alpha1/contour_types.go
```diff
@@ -229,6 +229,41 @@ type EnvoyNetworkPublishing struct {
 	// +kubebuilder:validation:MaxItems=2
 	// +kubebuilder:default={{name: http, portNumber: 8080}, {name: https, portNumber: 8443}}
 	ContainerPorts []ContainerPort `json:"containerPorts,omitempty"`
+
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// +optional
+	Tolerations []Toleration `json:"tolerations,omitempty"`
```
skriss, line 237: I'm not sure these belong under `NetworkPublishing`, we may want to move them elsewhere in the API.

skriss (commented):
Thanks for the PR @amdonov! I definitely think this feature makes sense to add. xref #362 and cc @danehans for thoughts as well

api/v1alpha1/contour_types.go
danehans: I prefer using `corev1.Toleration` instead of supporting a custom type.

amdonov:
Sounds good. I'll close this PR and open two different ones following your
recommendations.

amdonov:
Made custom types for ContourNodePlacement and EnvoyNodePlacement as requested. Can't think of any fields that would differ between the two. Recommend using a single type if you agree.

danehans:
@amdonov I started providing feedback, but I stopped since I felt it was becoming confusing. Instead, I created a [commit](https://github.com/danehans/contour-operator/commit/c383af4d5429a6781b867eac2e163b6ffb05478f) that details the suggested changes, PTAL when you have a moment. Tests are still needed and I want to make sure you and others agree to use `node-role.kubernetes.io/worker: ""` as the default node selector.

> Made custom types for ContourNodePlacement and EnvoyNodePlacement as requested. Can't think of any fields that would differ between the two. Recommend using a single type if you agree.

Although the two types are the same today, they may differ in the future. Therefore I prefer defining separate types.

cc: @stevesloka @skriss @sunjayBhatia @youngnick

amdonov:
@danehans I'd recommend not setting a default node selector. I checked labels on a couple of the k8s distros that I use frequently. For the control plane nodes one had node-role.kubernetes.io/master=true, and the other had node-role.kubernetes.io/master="". However, the worker nodes weren't labeled in either case. I'm afraid the pods wouldn't get scheduled and could lead to confusion.

danehans:
> I'd recommend not setting a default node selector.

@amdonov thanks for looking into this. I was unsure how distros handled the node roles by default. With that said, here is the [updated commit](https://github.com/danehans/contour-operator/commit/b118f5459c11bb2c66bceaaf836ce7be7e56f295). Again, unit tests are still required.

amdonov:
I'm done pending further feedback.

examples/operator/operator.yaml
```diff
@@ -403,6 +403,30 @@ spec:
                               information specific to the underlying infrastructure
                               provider.
                             properties:
+                              aws:
```
danehans, line 406: __Note__: This is leftover from https://github.com/projectcontour/contour-operator/pull/336

danehans (approved):

api/v1alpha1/helpers.go
```diff
@@ -51,3 +51,47 @@ func (c *Contour) IsFinalized() bool {
...
+	if c.Spec.NodePlacement != nil &&
+		c.Spec.NodePlacement.Contour != nil &&
+		len(c.Spec.NodePlacement.Contour.Tolerations) > 0 {
+		return true
+	}
+
+	return false
+}
+
+// EnvoyNodeSelectorExists returns true if a nodeSelector is specified for Contour.
```
skriss, line 77:
```suggestion
// EnvoyNodeSelectorExists returns true if a nodeSelector is specified for Envoy.
```

api/v1alpha1/helpers.go
```diff
@@ -51,3 +51,47 @@ func (c *Contour) IsFinalized() bool {
...
+	if c.Spec.NodePlacement != nil &&
+		c.Spec.NodePlacement.Envoy != nil &&
+		c.Spec.NodePlacement.Envoy.NodeSelector != nil {
+		return true
+	}
+
+	return false
+}
+
+// EnvoyTolerationsExist returns true if a nodeSelector is specified for Envoy.
```
skriss, line 88:
```suggestion
// EnvoyTolerationsExist returns true if tolerations are set for Envoy.
```

skriss (changes requested):
couple godoc nits but otherwise LGTM, thanks @amdonov!

skriss (approved):
LGTM. DCO bot is unhappy with the GH suggestion commits but we can address that when we squash+merge.

PR #374, opened 2021-05-28
Bump kind cli and images

sunjayBhatia:
See release notes:
https://github.com/kubernetes-sigs/kind/releases/tag/v0.11.1

### DIFF
--- a/.github/workflows/prbuild.yaml
+++ b/.github/workflows/prbuild.yaml
@@ -43,11 +43,11 @@ jobs:
         # image to use) for each kubernetes_version value.
         include:
           - kubernetes_version: "kubernetes:latest"
-            node_image: "docker.io/kindest/node:v1.21.1"
+            node_image: "docker.io/kindest/node:v1.21.1@sha256:69860bda5563ac81e3c0057d654b5253219618a22ec3a346306239bba8cfa1a6"
           - kubernetes_version: "kubernetes:n-1"
-            node_image: "docker.io/kindest/node:v1.20.7"
+            node_image: "docker.io/kindest/node:v1.20.7@sha256:cbeaf907fc78ac97ce7b625e4bf0de16e3ea725daf6b04f930bd14c67c671ff9"
           - kubernetes_version: "kubernetes:n-2"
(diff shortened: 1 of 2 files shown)


skriss (approved):

.github/workflows/prbuild.yaml
```diff
@@ -78,9 +78,9 @@ jobs:
           - kubernetes_version: "kubernetes:latest"
             node_image: "docker.io/kindest/node:v1.21.1"
           - kubernetes_version: "kubernetes:n-1"
-            node_image: "docker.io/kindest/node:v1.20.2"
+            node_image: "docker.io/kindest/node:v1.20.7"
           - kubernetes_version: "kubernetes:n-2"
-            node_image: "docker.io/kindest/node:v1.19.7"
+            node_image: "docker.io/kindest/node:v1.19.11"
```
danehans, line 83: @sunjayBhatia per the [release notes](https://github.com/kubernetes-sigs/kind/releases/tag/v0.11.1), it's suggested to include the digest in the image. Thoughts?

danehans (changes requested):

.github/workflows/prbuild.yaml
sunjayBhatia: yeah we totally could do so, i didnt b/c we had been using just the tag with no issue until now

looks like this originated here: https://github.com/kubernetes-sigs/kind/releases/tag/v0.3.0 probably b/c of incompatibilities between the CLI and image structure, but maybe not an issue since we stay on the latest CLI as well as images?

danehans: The above link states:
> We hope not to make further breaking changes to the images, but strongly
recommend pinning images by sha256, the default image is pinned to a particular sha.

Could we get in a situation where we troubleshoot ghosts due to unknowingly using a new image (since we're not specifying a digest)? cc: @stevesloka @skriss

stevesloka: Agree it's probably not an issue, but wouldn't hurt to use the full image just to be safe.

danehans (approved):

Issue #375, opened 2021-06-02
update example operator deployment to support arm64


vchrisb:
The image currently used for `kube-rbac-proxy` is `gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0`.
https://github.com/projectcontour/contour-operator/blob/main/examples/operator/operator.yaml#L6531 

This image is published by kubebuilder project rather than the kube-rbac-proxy project and is not multi-arch.

kube-rbac-proxy project is publishing to `quay.io/brancz/kube-rbac-proxy` and these images are multi-arch.

Updating the deployment to use kube-rbac-proxy v0.6.0+ from `quay.io/brancz/kube-rbac-proxy` would allow using the operator on arm64.

danehans:
Fixed by https://github.com/projectcontour/contour-operator/pull/376

PR #376, opened 2021-06-02
update kube-rbac-proxy image version to support arm64

vchrisb:
This is addressing #375.

Contour and Envoy are supported on arm64 and do publish multi-arch images.
The contour-operator does leverage `kube-rbac-proxy` image `v0.5.0` published by kubebuilder project. 
These images are not multi-arch.

The `kube-rbac-proxy` project does publish to `quay.io/brancz/kube-rbac-proxy` and these images are build for multi-arch since `v0.6.0`.
This PR also use latest `v0.10.0`, which is now based on `gcr.io/distroless/static` and running nonroot

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -11,7 +11,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: quay.io/brancz/kube-rbac-proxy:v0.10.0
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
         args:
         - "--secure-listen-address=0.0.0.0:8443"
         - "--upstream=http://127.0.0.1:8080/"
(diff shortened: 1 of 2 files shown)


sunjayBhatia:
Hey @vchrisb to get DCO to pass could you amend your commit(s) with the `--signoff` flag? Thanks!

vchrisb:
@sunjayBhatia should be fine now 👍

danehans (approved):

sunjayBhatia (commented):
Do we have a link to documentation about the image being hosted in the registry it is? Given it is not a k8s org project, just good to do some due diligence and inspect the image itself

danehans:
@vchrisb as a follow-on to the above comment from @sunjayBhatia, do you have a reference to other k8s projects that use the image/repo?

vchrisb:
good point @danehans 
I did some more digging and it is actually enough to use `gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0`.
They updated the build environment to support multi-arch.
So a version upgrade to `v0.8.0`is enough to get multi-arch support and leverage a more secure container.
I will update the PR.

skriss (approved):
LGTM, thanks @vchrisb!

Issue #263, opened 2021-03-15
Decision: Simplify Namespace Management for Contour API


danehans:
**Please describe the problem you have**
The Contour API exposes the ability to manage the namespace of the operands, i.e. Contour deployment. The `NamespaceSpec` type contains `name`, the name of the operand namespace to manage and `removeOnDeletion`, which indicates whether to remove the namespace when deleting the Contour. I suggest simplifying this schema and removing `removeOnDeletion`. The operator already contains the logic to not remove the managed operand namespace for the following conditions:
```

// 1. The Contour namespace is "default", "kube-system" or the
//    contour-operator's namespace.
//
// 2. Another Contour exists in the namespace.
//
// 3. The namespace does not contain the Contour owning label.
```

When https://github.com/projectcontour/contour-operator/issues/260 is fixed, the operator will not delete the gateway namespace if pods exist in the namespace and any pod surfaces the "Ready=True" status condition. The same approach can be taken for the contour controller. These mechanisms should provide adequate safeguards for removing the operand namespace. The new `ContourSpec` would look like:
```
// ContourSpec defines the desired state of Contour.
type ContourSpec struct {
	// Namespace defines the name of the namespace to provision Contour.
	//
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=253
	// +kubebuilder:default=projectcontour
	Namespace string `json:"namespace,omitempty"`
```

Thoughts @youngnick @stevesloka

stevesloka:
I think this extra logic is great @danehans!

youngnick:
Yup, agreed from me as well.

danehans:
xref Slack discussion regarding proposed approach: https://kubernetes.slack.com/archives/C8XRH2R4J/p1622848335153000

PR #382, opened 2021-06-09
Simplifies Namespace Management

danehans:
Removes namespace management from `contour.spec`. The operator will now provision operands, i.e. Envoy DaemonSet, in the meta namespace of the Contour.

- `api/v1alpha1/contour_types.go`:  Removes namespace management from Contour spec.
- `config/rbac/role.yaml`: Removes namespace RBAC since the operator no longer manages namespaces.
- `examples/*`: Updates namespaces, RBAC, Contour CRD, etc. in example manifests.
- `hack/release/make-release-tag.sh `: Updates e2e to set the release image for the operator upgrade test case.
- `internal/objects/*`: Updates resources to derive the namespace from Contour meta namespace instead of spec namespace.
- `internal/objects/namespace/*`: Removes the namespace package.
- `internal/operator/controller/*`: Removes namespace management from controller reconcilitation. 
- `internal/operator/suite_test.go`: Removes spec namespace integration test cases.
- `pkg/validation/*`: Updates Contour validation to use meta namespace instead of spec namespace.
- `test/e2e/operator_test.go`: Updates tests to use Contour meta namespace instead of spec namespace. Updates the operator upgrade tests to use a specific release instead of "latest". Adds downgrading/upgrading the CRDs as part of the operator upgrade test.
-  `test/e2e/utils.go`: Removes the function for deleting managed namespaces.

Fixes https://github.com/projectcontour/contour-operator/issues/263


api/v1alpha1/contour_types.go
```diff
@@ -66,16 +66,6 @@ type ContourSpec struct {
...
-	// Namespace defines the schema of a Contour namespace. See each field for
-	// additional details. Namespace name should be the same namespace as the
-	// Gateway when GatewayClassRef is set.
-	//
-	// TODO [danehans]: Ignore Namespace when GatewayClassRef is set.
-	// xref: https://github.com/projectcontour/contour-operator/issues/212
-	//
-	// +kubebuilder:default={name: "projectcontour", removeOnDeletion: false}
-	Namespace NamespaceSpec `json:"namespace,omitempty"`
-
```
skriss, line 78: I believe that, if we're following upstream versioning standards, removing this field requires an API bump to `v1alpha2`:

> (from https://kubernetes.io/docs/reference/using-api/deprecation-policy/ -- Rule 1) Once an API element has been added to an API group at a particular version, it can not be removed from that version or have its behavior significantly changed, regardless of track.

internal/objects/daemonset/daemonset_test.go
```diff
@@ -123,8 +123,6 @@ func TestDesiredDaemonSet(t *testing.T) {
 	cfg := objcontour.Config{
 		Name:        name,
 		Namespace:   fmt.Sprintf("%s-ns", name),
-		SpecNs:      "projectcontour",
-		RemoveNs:    false,
```
skriss, line 127: it'd be nice to have this test verify that the DS gets created in `cfg.Namespace`. Similarly for the other `TestDesired*` tests.

internal/objects/rolebinding/role_binding_test.go
```diff
@@ -69,12 +69,10 @@ func TestDesiredRoleBinding(t *testing.T) {
 	cfg := objcontour.Config{
 		Name:        name,
 		Namespace:   fmt.Sprintf("%s-ns", name),
-		SpecNs:      "projectcontour",
-		RemoveNs:    false,
 		NetworkType: operatorv1alpha1.LoadBalancerServicePublishingType,
 	}
 	cntr := objcontour.New(cfg)
-	cntr.Spec.Namespace.Name = "test-rb-ns"
+	cntr.Namespace = "test-rb-ns"
```
skriss, line 75: can't we remove this line and set the NS directly on L71 instead?

internal/operator/operator.go
```diff
@@ -59,6 +59,7 @@ type Operator struct {
 // +kubebuilder:rbac:groups=operator.projectcontour.io,resources=contours/status,verbs=get;update;patch
 // cert-gen needs create/update secrets.
 // +kubebuilder:rbac:groups="",resources=namespaces;secrets;serviceaccounts;services,verbs=get;list;watch;delete;create;update
+// +kubebuilder:rbac:groups="",resources=secrets;serviceaccounts;services,verbs=get;list;watch;delete;create;update
```
skriss, line 62: these two lines are identical except the first also includes `namespaces`, should be de-duped.

test/e2e/operator_test.go
```diff
@@ -35,7 +36,11 @@ import (
 	gatewayv1alpha1 "sigs.k8s.io/gateway-api/apis/v1alpha1"
 )
 
-const timeout = 3 * time.Minute
+const (
+	timeout = 3 * time.Minute
+	// latestRelease is the "from" version used to test operator upgrades.
+	latestRelease = "v1.16.0"
```
skriss, line 42: is this now needed because the upgrade test is also upgrading CRDs, so using "latest" won't work since you need to know the specific version to pull from GH? Just making sure I'm following the change.

skriss (changes requested):
This looks pretty good @danehans, had a handful of comments

danehans:
Converting to draft since this is a breaking change that requires an API bump, so https://github.com/projectcontour/contour-operator/issues/386 needs to be resolved first.

api/v1alpha1/contour_types.go
danehans: @skriss thanks for pointing this out. I marked this PR as a draft and created https://github.com/projectcontour/contour-operator/issues/386 to discuss the API bump details.

danehans:
@skriss commit `5316936` resolves all your feedback except the breaking API change. I'm continuing to keep this PR as a draft until  #386 is resolved.

danehans:
Closing since https://github.com/projectcontour/contour-operator/issues/386 is required. After https://github.com/projectcontour/contour-operator/issues/386 is resolved, this PR should be reopened.

PR #390, opened 2021-06-22
Add example contour deployment with internal-elb

erismaster:
Adds an example contour custom resoruce object that would provision with
an internal-facing load balancer.


danehans (approved):

skriss (approved):
LGTM, thanks @erismaster!

Issue #112, opened 2020-11-13
Configurable Security Context


danehans:
**Please describe the problem you have**
Contour and the certgen job do not run on OpenShift using the default security context:
```
securityContext:
  runAsNonRoot: true
  runAsUser: 65534
  runAsGroup: 65534
```

Failure status condition of deploy/contour:
```
  - lastTransitionTime: "2020-11-13T21:44:15Z"
    lastUpdateTime: "2020-11-13T21:44:15Z"
    message: 'pods "contour-5475898957-" is forbidden: unable to validate against
      any security context constraint: [spec.containers[0].securityContext.runAsUser:
      Invalid value: 65534: must be in the ranges: [1000590000, 1000599999]]'
    reason: FailedCreate
    status: "True"
    type: ReplicaFailure
```

Either make `securityContext` configurable or introduce provider-specific extensions to `Contour` so this can be configured automatically.

/cc @jpeach @Miciah

youngnick:
I think in this case, that the securityContext of the Contour and Envoy pods should be configurable, maybe just with a single "userID" or similar setting.

PR #393, opened 2021-06-24
Running with UID 101 and GID 101 when using high ports

amdonov:
Our security team really hates us running any containers as root. Haven't done any testing on this, but it seems like it's reasonable to run envoy with an alternate UID when not attempting to bind to 80 and 443.

Interested in feedback both on behavior and approach. We're using the public envoy image, the default for the operator. Based upon a quick read it looks like the envoy user and group in that image has UID/GID of 101.

### DIFF
--- a/internal/objects/daemonset/daemonset.go
+++ b/internal/objects/daemonset/daemonset.go
@@ -146,8 +146,9 @@ func DesiredDaemonSet(contour *operatorv1alpha1.Contour, contourImage, envoyImag
 	// run as 1001 if not using any low ports
 	securityContext := &corev1.PodSecurityContext{}
 	if minPort > 1024 {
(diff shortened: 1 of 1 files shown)


danehans:
xref: https://github.com/projectcontour/contour-operator/issues/112

amdonov:
Don't know how I missed that existing ticket. If you want to allow users to set security context parameters as part of the CRD, I'll implement it that way. Was trying to improve security without requiring additional configuration.

youngnick:
I agree that this being configurable in the CRD would be ideal, we can default to 65534 - I think that it's better to default to nonroot, and need to change to use low ports.

amdonov:
@dmorgan81 is implementing the suggested approach in #398

Issue #392, opened 2021-06-23
Migrate to Kubebuilder v3


danehans:
**Please describe the problem you have**
The operator was bootstrapped using kubebuilder v2 and should consider migrating to v3. Reference the [migration guide](https://book.kubebuilder.io/migration/v2vsv3.html) for additional details.

cc: @BostjanBozic

BostjanBozic:
I have been fiddling around a bit, checking how to approach this. My question is, should we try and follow new scaffolded structure/parameters as much as possible or do we just adjust `PROJECT` file? Some examples:
* scaffolded manager deployment manifest has `health-probe-bind-address` flag, should we add it to `contour-operator.go`?
* `Makefile` as quite some changes, e.g. testing is now performed on a fly using temp downloaded binaries for etcd
* container name of operator is `manager`, in our case it is `contour-operator`

I checked codebase and apart from `PROJECT` file and some of the naming/structural changes, everything is already upgraded (e.g. `Reconcile` function is already inline with v3, CRDs are already on `apiextensions.k8s.io/v1`).

youngnick:
It seems like if we want to be able to use kubebuilder's CLI tooling, we will need to follow the scaffolded structure as much as possible. That seems like a slightly better plan to me.

PR #396, opened 2021-06-28
Migrate to Kubebuilder v3 schema

BostjanBozic:
Updates project schema to Kubebuilder v3 compliant schema. Following changes have been performed:
* `PROJECT` file is now Kubebuilder v3 compliant
* Initial comment updates to `api/v1alpha1`
* `examples` folder has been moved to `config/samples` to follow Kubebuilder v3 scaffolding
* `Makefile` has been revamped - added `help`, reshuffled command order, updated `test` target to use new `setup-env` binary (could also be used for https://github.com/projectcontour/contour/issues/3832)
* updated `Kustomize` templating based on Kubebuilder scaffolding. `namePrefix` is not used, since it changes deployment name to `contour-operator-contour-operator`, which breaks `test/e2e/operator_test.go` (same string used for deployment name and container name). If wanted, we can sort this out as well.
* Specify user in container image (change from `nonroot` to UID `65532` (default by Kubebuilder)
* `controller-runtime` update from `v0.9.0-beta0` to `v0.9.2`

There are quite some minor tweaks, mostly to naming conventions, so it would be great if someone else can take a look as well and provide some feedback what we want to keep "the old way" and what in "kubebuilder v3" way.

Regarding failing `test-e2e`:
* Test is failing on updating operator image. I believe reason for this failure is that `latest` tag still includes `nonroot` user and since new deployment includes parameter `securityContext.allowPrivilegeEscalation: false`, it fails on deploying `latest` version. As `waitForImage` is waiting for all pods with label `control-plane: contour-operator` to have latest image, it reaches timeout (old pod is never terminated, as deletion is not triggered, since loop is waiting for new pod to become available).

Updates: https://github.com/projectcontour/contour-operator/issues/392

cc: @danehans @youngnick 


sunjayBhatia:
Initial impression from me is we might want to break this up into smaller more digestible pieces to review since the number of changes here is pretty big for a single PR

sunjayBhatia:
Also consider waiting to merge all this to after the upcoming release since this is a large change (size maybe not actual complexity)

BostjanBozic:
@sunjayBhatia thanks for looking into it. I am up for waiting to do this after next release. Regarding breaking this into smaller pieces, I would try to keep most of changes in single PR (most of the changes are reshuffling file location and comments), but maybe we could have a separate one for `Makefile` changes.

BostjanBozic:
@sunjayBhatia Do you think it would be time to look into this now that v1.17.0 is released? I also saw there were updates to `controller-runtime`, so I would maybe wait until this is pushed in kubebuilder and then review additional changes this creates?

skriss:
This is stale, closing out.

Issue #378, opened 2021-06-02
Support envoy pod template with basic fields


r0bj:
**Please describe the problem you have**
[A clear, concise, description of the problem you are facing. What is the _problem_ that feature X would solve for you?]

It would be useful to be able to specify envoy pod template with basic fields like:
- labels
- annotations
- securityContext
etc.

In some environments/clusters there are specific requirements for pod labels (e.g. mandatory label team, owner, contact, etc.). Other use case is for specific annotation which can be used to define pod log destination or other parameters.

danehans:
@r0bj in my experience template-based approaches become very difficult to manage. Instead of exposing a "pod template" through the API, I would like to address each of the above bullets in a specific manner. Using labels as an example, we can have the operator append labels defined in the Contour to the operands:
```
apiVersion: operator.projectcontour.io/v1alpha1
kind: Contour
metadata:
  name: contour-sample
  namespace: projectcontour
  labels:
    environment: production
    app: foo
spec:
...

# Results in operands having the labels:
apiVersion: apps/v1
kind: DeamonSet
metadata:
  name: envoy
  namespace: projectcontour
  labels:
    environment: production
    app: foo
 ...
```

A similar approach can be taken for custom annotations.

We already have an issue for managing SCCs: https://github.com/projectcontour/contour-operator/issues/112.

What do you think about these approaches?

r0bj:
Sure, it would be less flexible but it would work. Partly my inspiration was prometheus-operator where you can define `podMetadata` and also separate fields for: `affinity`, `tolerations`, `securityContext`, `resources`, etc. Similarly `cert-manager` and `ClusterIssuer`: you can configure `podTemplate` with `labels`, `annotations`, `nodeSelector`, etc.

danehans:
xref https://github.com/projectcontour/contour-operator/pull/373 for tolerations/nodeSelector support of Envoy pods.

PR #398, opened 2021-07-06
Custom SecurityContexts for envoy DaemonSet and contour Deployment

### LINKED ISSUE #112: Configurable Security Context

dmorgan81:
Allow users to specify a security context for both envoy daemonsets and contour deployments. The default value is unprivileged and is equivalent to the following:

```
contourSecurityContext:
  runAsUser: 65534
  runAsGroup: 65534
  runAsNonRoot: true
```

Note that this does change the default behavior for envoy daemonsets. Currently an envoy daemonset has an empty security context, meaning it will run with the user/group specified in the envoy image, which is commonly root. Switching to a non-root user by default will prevent the envoy container from binding to a low port. Existing setups will need to be updated by adding a `envoySecurityContext` entry with the appropriate settings if they need to bind to a low port.

Fixes #112
Updates #378


dmorgan81:
I've been trying to run the e2e tests locally but I'm getting inconsistent results, even on the main branch:

- TestContourNodePortService often hangs.
- TestGateway and TestGatewayOwnership sometimes fail. The contour pods associated with these tests are stuck in a crash loop with the only logging being `contour: error: invalid Contour configuration: invalid Gateway parameters specified: controllerName required, try --help`

I just added a change to the e2e tests to use an empty PodSecurityContext for the contour pods since port 80 appears to be in the mix.  However, I'm unable to get through the entire e2e test suite because of the above two issues.

sunjayBhatia:
> I've been trying to run the e2e tests locally but I'm getting inconsistent results, even on the main branch:
> 
> * TestContourNodePortService often hangs.
> * TestGateway and TestGatewayOwnership sometimes fail. The contour pods associated with these tests are stuck in a crash loop with the only logging being `contour: error: invalid Contour configuration: invalid Gateway parameters specified: controllerName required, try --help`
> 
> I just added a change to the e2e tests to use an empty PodSecurityContext for the contour pods since port 80 appears to be in the mix. However, I'm unable to get through the entire e2e test suite because of the above two issues.

ah, this is probably not related to your change at all, i think this is related to this change in Contour being merged: https://github.com/projectcontour/contour/commit/9e98a7284e680a3f71c14d35195588175e7605a6

We need to fix this up on main and then you should be able to merge main and your change go green

internal/objects/deployment/deployment.go
```diff
@@ -304,7 +304,6 @@ func DesiredDeployment(contour *operatorv1alpha1.Contour, image string) *appsv1.
 					ServiceAccountName:            objutil.ContourRbacName,
 					RestartPolicy:                 corev1.RestartPolicyAlways,
 					SchedulerName:                 "default-scheduler",
-					SecurityContext:               objutil.NewUnprivilegedPodSecurity(),
```
nak3, line 307: I think we don't need to move this line.
If `contour.ContourSecurityContextExists()` is `true`, we just overrides `deploy.Spec.Template.Spec.SecurityContext`.

api/v1alpha1/contour_types.go
```diff
@@ -111,6 +111,16 @@ type ContourSpec struct {
 	//
 	// +optional
 	NodePlacement *NodePlacement `json:"nodePlacement,omitempty"`
+
+	// ContourSecurityContext defines a PodSecurityContext for Contour pods.
+	//
+	// +optional
+	ContourSecurityContext *corev1.PodSecurityContext `json:"contourSecurityContext,omitempty"`
+
+	//EnvoySecurityContext defines a PodSecurityContext for Envoy pods.
```
nak3, line 120:
```suggestion
	// EnvoySecurityContext defines a PodSecurityContext for Envoy pods.
```

nit

api/v1alpha1/helpers.go
```diff
@@ -95,3 +95,11 @@ func (c *Contour) EnvoyTolerationsExist() bool {
 
 	return false
 }
+
```
nak3, line 98: I think `ContourSecurityContextExists` and `EnvoySecurityContextExists` should have a comment line as these are public functions. I wonder why lint check does not produce an error :thinking:

nak3 (commented):
I also wanted this feature! Thank you :+1:

internal/objects/deployment/deployment.go
dmorgan81: It might be premature optimization but I was hoping to avoid an unnecessary allocation.

sunjayBhatia:
@dmorgan81 I think this PR might be placed on a lower priority to merge at the moment since the Operator/Contour relationship is going through a rework as part of our Gateway API support work. Apologies for that, but we will try to get to this soon!

sunjayBhatia (commented):
Added https://github.com/projectcontour/contour/issues/4039 and https://github.com/projectcontour/contour/issues/4040 to track we want to make sure this is captured in the new world where Contour can manage the Envoy daemonset etc.

arjunrn:
@dmorgan81 The jobs created by the operator will also have to use the custom security context. For [reference](https://github.com/openshift/contour-operator/pull/10).

skriss:
This is stale, closing out.

Issue #400, opened 2021-07-07
Operator should configure Contour with gateway controller name


sunjayBhatia:
Contour now requires the gateway controller string to be set in its config when using Gateway API

Currently, Contours deployed by the operator fail with:

```
contour: error: invalid Contour configuration: invalid Gateway parameters specified: controllerName required, try --help
```

We also need to see if we need to start removing some of the gateway/gateway class controller logic in the operator now that Contour itself has a gateway and gateway class controller

youngnick:
Investigating this, I have a patch that fixes this problem, but a further change in Contour has made GatewayClasses with non-nil parametersRef invalid, this should be ignored. Working on a patch for that now, and I'll log an issue in Contour to track that work.

Once those two things are fixed, I think we will need to cut a .1 release to fix this, for both products.

PR #401, opened 2021-07-08
Fixes for Contour now reconciling GatewayClasses and Gateways

sunjayBhatia:
Now that Contour reconciles and sets status on GatewayClass and Gateway
resources, this is the first step to clean up the operator to remove
bits that conflict with Contour itself.

- Pass controller name string to Contour in ConfigMap to ensure Contour
is properly configured for Gateway API
- Remove status updating/checking on Gateway API resources
  - Don't set GatewayClass status or expect it to be admitted by
Operator
  - Don't set Gateway status or expect it to be set in tests
  - When reconciling Contour CRD, do not check if referenced
GatewayClass is admitted
- Adds test to make sure multiple Contours with Gateway API work
  - Currently this is a bit broken and not how Gateway API *should* work
  - Due to the fact that the Operator only reconciles 1 GatewayClass and passes that along to Contour
  - This will be cleaned up in next release

Fixes: #400

### DIFF
--- a/internal/objects/configmap/configmap.go
+++ b/internal/objects/configmap/configmap.go
@@ -46,7 +46,7 @@ var contourCfgTemplate = template.Must(template.New("contour.yaml").Parse(`
 #   determine which XDS Server implementation to utilize in Contour.
 #   xds-server-type: contour
 #
-# Specify the service-apis Gateway Contour should watch.{{if and .GatewayName .GatewayNamespace .GatewayControllerName }}
+# Specify the Gateway API configuration.{{if and .GatewayName .GatewayNamespace .GatewayControllerName }}
 gateway:
   controllerName: {{.GatewayControllerName}}
   name: {{.GatewayName}}
--- a/internal/objects/configmap/configmap.go
+++ b/internal/objects/configmap/configmap.go
@@ -46,8 +46,9 @@ var contourCfgTemplate = template.Must(template.New("contour.yaml").Parse(`
 #   determine which XDS Server implementation to utilize in Contour.
 #   xds-server-type: contour
 #
-# Specify the service-apis Gateway Contour should watch.{{if and .GatewayName .GatewayNamespace}}
+# Specify the service-apis Gateway Contour should watch.{{if and .GatewayName .GatewayNamespace .GatewayControllerName }}
 gateway:
+  controllerName: {{.GatewayControllerName}}
   name: {{.GatewayName}}
   namespace: {{.GatewayNamespace}}{{else}}
 # gateway:
@@ -164,6 +165,9 @@ type contourConfig struct {
 	GatewayNamespace string
 	// GatewayName is the Gateway name Contour should watch.
 	GatewayName string
+	// GatewayControllerName is the name of the controller that should
+	// reconcile GatewayClasses and associated Gateways.
+	GatewayControllerName string
 }
 
 // NewConfig returns a Config with default fields set.
@@ -181,13 +185,14 @@ func NewCfgForContour(contour *operatorv1alpha1.Contour) *Config {
 }
 
 // NewCfgForGateway returns a ConfigMap Config with default fields set for gw.
-func NewCfgForGateway(gw *gatewayv1alpha1.Gateway) *Config {
+func NewCfgForGateway(gw *gatewayv1alpha1.Gateway, controllerName string) *Config {
 	cfg := NewConfig()
 	cfg.Namespace = gw.Namespace
 	labels := objgw.OwnerLabels(gw)
 	cfg.Labels = labels
 	cfg.Contour.GatewayNamespace = gw.Namespace
 	cfg.Contour.GatewayName = gw.Name
+	cfg.Contour.GatewayControllerName = controllerName
 	return cfg
 }
 
--- a/internal/objects/gateway/gateway.go
+++ b/internal/objects/gateway/gateway.go
@@ -78,6 +78,10 @@ func ContourForGateway(ctx context.Context, cli client.Client, gw *gatewayv1alph
 		return nil, nil
 	}
 
+	if gc.Spec.ParametersRef == nil {
+		return nil, fmt.Errorf("failed to get contour for gateway %s/%s and gatewayclass %s, no gatewayclass parameters reference", gw.Namespace, gw.Name, gc.Name)
+	}
+
 	cntr, err := objcontour.CurrentContour(ctx, cli, *gc.Spec.ParametersRef.Namespace, gc.Spec.ParametersRef.Name)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get contour for gateway %s/%s", gw.Namespace, gw.Name)
--- a/internal/operator/controller/gateway/controller.go
+++ b/internal/operator/controller/gateway/controller.go
@@ -28,7 +28,6 @@ import (
 	objjob "github.com/projectcontour/contour-operator/internal/objects/job"
 	objns "github.com/projectcontour/contour-operator/internal/objects/namespace"
 	objsvc "github.com/projectcontour/contour-operator/internal/objects/service"
-	"github.com/projectcontour/contour-operator/internal/operator/status"
 	retryable "github.com/projectcontour/contour-operator/internal/retryableerror"
 	"github.com/projectcontour/contour-operator/pkg/validation"
 
@@ -166,9 +165,6 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 						cntr.Namespace, cntr.Name, err))
 				}
 			}
-			if err := status.SyncGateway(ctx, r.client, gw); err != nil {
-				errs = append(errs, fmt.Errorf("failed to sync status for gateway %s/%s: %w", gw.Namespace, gw.Name, err))
-			}
 		default:
 			// Before doing anything with the gateway, ensure it has a finalizer
 			// so it can cleaned-up later.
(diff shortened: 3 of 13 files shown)


sunjayBhatia:
All green besides the codecov check on the patch diff

internal/objects/configmap/configmap.go
```diff
@@ -46,8 +46,9 @@ var contourCfgTemplate = template.Must(template.New("contour.yaml").Parse(`
 #   determine which XDS Server implementation to utilize in Contour.
 #   xds-server-type: contour
 #
-# Specify the service-apis Gateway Contour should watch.{{if and .GatewayName .GatewayNamespace}}
+# Specify the service-apis Gateway Contour should watch.{{if and .GatewayName .GatewayNamespace .GatewayControllerName }}
```
stevesloka, line 49: nit: s/service-apis/gateway-apis/

internal/objects/gateway/gateway.go
```diff
@@ -78,6 +78,10 @@ func ContourForGateway(ctx context.Context, cli client.Client, gw *gatewayv1alph
 		return nil, nil
 	}
 
+	if gc.Spec.ParametersRef == nil {
```
stevesloka, line 81: Something to make sure we're ok with. In the past, we've allowed an empty spec and used a set of defaults, but now it's required. 

The problem is knowing if we're going to use GatewayAPI's in contour or not. Today that's defined by specifying the `gateway` bits in the config file.

internal/operator/controller/gateway/controller.go
```diff
@@ -215,7 +211,7 @@ func (r *reconciler) ensureGateway(ctx context.Context, gw *gatewayv1alpha1.Gate
 	}
 
 	// configmap error/logging messages are different, hence not using handleResult
-	if err := objcm.Ensure(ctx, cli, objcm.NewCfgForGateway(gw)); err != nil {
+	if err := objcm.Ensure(ctx, cli, objcm.NewCfgForGateway(gw, operatorv1alpha1.GatewayClassControllerRef)); err != nil {
```
stevesloka, line 214: Should the `NewCfgForGateway` use the passed in controllerName?

stevesloka (changes requested):
Just a couple comments and one nit.

internal/objects/configmap/configmap.go
sunjayBhatia: 👍🏽 made it match the Contour configmap comment

internal/objects/gateway/gateway.go
sunjayBhatia: yeah the validation in the operator doesn't let the parametersref be nil, so at this point in the code we just assumed it was set, which causes panics (before this PR this was guarded by the check that the operator had admitted the gw class, which can only happen if there is a parametersref)

i'm thinking this stopgap is ok since we're going to be removing the controllers altogether from the operator soon and in practice if you were using the operator + gw api then you had to make a gw class with parametersref already

internal/operator/controller/gateway/controller.go
sunjayBhatia: i used this shortcut b/c the operator already expects a gw class with that controller name string (otherwise it doesnt reconcile it) and the operator checks that the gw class referenced by the gateway has that same string as well

I would say maybe as part of our next refactors to remove the gw controllers for this we should do the change to the Contour CRD Gateway ref field and rename/repurpose it to be a controller string (I'll add that to the proposal doc)

internal/objects/gateway/gateway.go
stevesloka: We should also update the example so pages like this still work: https://projectcontour.io/guides/gateway-api/

internal/operator/controller/gateway/controller.go
stevesloka: It's just that if you have 2 instances running in a cluster, they'd end up with the same controllerName wouldn't they? That wouldn't allow Contour to process since it'd find duplicates.

sunjayBhatia: so with:
- gatewayclass name `gc1` (controller string `projectcontour.io/contour-operator`)
- gateway `ns1/gw1` referring to `gc1`
- gatewayclass name `gc2` (controller string `projectcontour.io/contour-operator`)
- gateway `ns2/gw2` referring to `gc2`
what happens in the operator:
- sees `gc1` and reconciles, its got the correct controller string (no longer sets status)
- sees `gc2` and reconciles, its got the correct controller string (no longer sets status)
- sees `gw1` and creates a `ns1/contour1` for it, passing `projectcontour.io/contour-operator` and `ns1/gw1`
- sees `gw2` and creates a `ns2/contour2` for it, passing `projectcontour.io/contour-operator` and `ns2gw2`
what happens in contour controller:
- `ns1/contour1` and `ns2/contour2` sees `gc1` and and both try to set status to admitted (this is weird/problematic)
- `ns1/contour1` and `ns2/contour2` sees `gc2` and also both try to set status to admitted? (we have't solved multiple gatewayclasses yet and this is weird/problematic)
- `ns1/contour1` sees `ns1/gw1` and reconciles
- `ns2/contour2` sees `ns2/gw2` and reconciles

Seems like some options are:
- operator doesn't hard code the controller string it looks for, should be configurable by user (repurpose or rename gateway ref like you said above)
  - Contour should also implement conflict handling of multiple gatewayclasses with the same controller string
- ditch the idea of using the controller string and just use gatewayclass names in Contour
- regardless of the above, the Operator should not allow users to create multiple contour instances referencing the *same* gatewayclass, by controller string or gatewayclass name (since I dont think that really makes sense)

internal/objects/gateway/gateway.go
sunjayBhatia: I think these work currently? gotta try it again but looks like the gwclasses referenced in the examples properly reference a parametersref

sunjayBhatia: yeah looks like things are working ok with this change

internal/operator/controller/gateway/controller.go
sunjayBhatia: I don't know how different the scenario is now than what it was before, I think there were issues before with having multiple contours, def something to fix but we're helped out by the fact that contour does reconcile a gw class but just sets status, currently contour's gateway processing doesn't check the admitted status of the gw class a gw references

sunjayBhatia:
@skriss @youngnick this one is ready for review and should fix Operator CI

internal/operator/controller/gateway/controller.go
skriss: Could we (for now) allow the operator to process any GatewayClass with a controller string starting with `projectcontour.io/`? And then if a user wants multiple Contours, they can use different controllerName strings starting with `projectcontour.io/` for each GatewayClass? Not sure if that would break things or not but maybe would work and allow >1 Contour to be created.

skriss:
Other than the thread about ways to support >1 Gateway-enabled Contour, this looks OK to me - I tried a basic test out locally and everything seemed to work.

internal/operator/controller/gateway/controller.go
sunjayBhatia: so just testing things out, multiple contours works though it is weird

skriss: Huh. It looks to me from the Contour code like the 2nd GatewayClass & Gateway created will take effect in *both* Contours because every time a new one comes in, the controllers `OnAdd` it to the cache/DAG, which overwrites whatever was already there - so if the 2nd GatewayClass/Gateway have identical specs to the first GatewayClass/Gateway, then yeah everything would appear to work, but otherwise the 1st Contour would end up with the 2nd Gateway's configuration too. Looking for an easy way to verify that..

sunjayBhatia: yeah, i may have to take back what i said about things working I had something broken locally, I think the multi-contour tests in the operator before weren't testing any requests actually work etc.

sunjayBhatia: yeah the multiple contour issue isn't going to be resolvable until the operator and Contour change significantly:
- operator no longer has a gw/gc controller
- operator doesn't allow users to create multiple Contours with the same controller string
- contour handles multiple gateways with the same gc name properly
- contour handles multiple gatewayclasses with the same controller string properly

sunjayBhatia: though checking the config dump, seems like the different contours are correctly also b/c the cache does filter on the named gateway that it is configured with

sunjayBhatia: ok added a test that verifies that 2 contours with different gateways can work still

sunjayBhatia:
@skriss @stevesloka @youngnick when you get a chance to review officially I think this should fix operator CI at least for 1.18

I'll open up some issues for what we need to do for the next release

sunjayBhatia:
release note should mention that the operator will no longer set status on GatewayClass/Gateways

internal/operator/controller/gateway/controller.go
skriss: Ah was wondering about https://github.com/projectcontour/contour-operator/issues/411#issuecomment-885070813 but saw you've thought about that already

skriss (approved):
OK, this seems to get us to a better place than we are today so LGTM. We've obviously got more work to do here. And we'll want to try to convey what works vs. doesn't in the release notes.

skriss:
woohoo ✅

skriss:
@stevesloka if you have a chance to review it'd be great to get this in so we can get the other PRs for 1.18 in too.

internal/operator/controller/gateway/controller.go
stevesloka: The operator should make this string unique to the instance it's deploying. So maybe something like `projectcontour.io/<namespace>/<name>` of the instance? 

Contour expects this string to be unique per GatewayClass/Gateway.

stevesloka (changes requested):

stevesloka (approved):

PR #402, opened 2021-07-08
add Slack notification if cronjob daily build fails

stevesloka:
Signed-off-by: Steve Sloka <slokas@vmware.com>


.github/workflows/cronbuild.yaml
```diff
@@ -0,0 +1,184 @@
...
+      - name: generate
+        run: |
+          make generate generate-contour-crds example manifests 
+          ./hack/actions/check-uncommitted-codegen.sh
+      - uses: act10ns/slack@v1
+        with:
+          status: ${{ job.status }}
+          steps: ${{ toJson(steps) }}
+          channel: '#contour-ci-notifications'
+        if: failure()
```
sunjayBhatia, line 183: might be able to keep it in the original file if we can check the branch as well? (e.g. `if failure && ${{ github.head_ref }} == main` or something)

might make it so the scheduled runs and the `main` runs both alert

stevesloka: yup that sounds good. Was looking for a way to not alert on each PR.

stevesloka: ok I think my final commit implements this properly. I just. did a run and did not get notified, so I think the next build would notify.

sunjayBhatia (approved):

sunjayBhatia:
lets try it out! worth adding to the `main` and release branch builds in the Contour repo as well I think

PR #405, opened 2021-07-08
Use github.ref instead of github.head_ref

sunjayBhatia:
the latter is only available on PRs


sunjayBhatia:
going to use admin powers to merge this

PR #410, opened 2021-07-14
internal: move config, controller, status packages

skriss:
Moves these three packages up one level, from under
internal/operator/ to directly under internal/. This
simplifies the package hierarchy.


skriss:
@stevesloka @youngnick @sunjayBhatia let me know if this seems like a reasonable change. Just looking to ramp back up on the operator and simplify some things where possible. Can definitely hold merging this until after any gateway-related changes, to avoid conflicts.

sunjayBhatia:
> @stevesloka @youngnick @sunjayBhatia let me know if this seems like a reasonable change. Just looking to ramp back up on the operator and simplify some things where possible. Can definitely hold merging this until after any gateway-related changes, to avoid conflicts.

sounds good to me, no need to have duplication/extra in the package paths if we dont need it

sunjayBhatia (approved):

stevesloka (approved):

skriss:
doesn't look like this should cause conflicts for other PRs now, ready for review/merge

PR #417, opened 2021-07-28
fixes for test-example

skriss:
In test-example, deletes Gateway API resources in the expected
order to avoid race conditions with finalizers that lead to
test flakes. Also adds a 30-minute timeout to the GitHub Actions
job.


sunjayBhatia (approved):
good call on the 30 min timeout as well!

stevesloka (approved):

PR #418, opened 2021-07-28
test/e2e: add a timeout to HTTP client

skriss:
Adds a 5-second timeout to the HTTP client to prevent
requests from hanging forever. Requests are wrapped in
a retry loop so timed-out requests will be retried.


skriss:
I believe this will address the issue encountered in https://github.com/projectcontour/contour-operator/runs/3185067010?check_suite_focus=true#step:5:310

youngnick (approved):
LGTM, nice work.

sunjayBhatia (approved):

PR #421, opened 2021-07-30
Add GatewayControllerName field to Contour CRD

sunjayBhatia:
- add field we need for properly configuring Contour when we remove
Gateway API controllers from Operator
- once we start using it, mark GatewayClassRef as deprecated
- GatewayClassRef can be removed in a future API revision/in Contour
config CRD when that comes about

### DIFF
--- a/api/v1alpha1/contour_types.go
+++ b/api/v1alpha1/contour_types.go
@@ -92,8 +92,9 @@ type ContourSpec struct {
 	GatewayClassRef *string `json:"gatewayClassRef,omitempty"`
 
 	// GatewayControllerName is used to determine which GatewayClass
-	// Contour reconciles. If unset, Contour will not reconcile Gateway
-	// API resources.
+	// Contour reconciles. The string takes the form of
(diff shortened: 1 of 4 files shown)


api/v1alpha1/contour_types.go
```diff
@@ -91,6 +91,14 @@ type ContourSpec struct {
 	// +optional
 	GatewayClassRef *string `json:"gatewayClassRef,omitempty"`
 
+	// GatewayControllerName is used to determine which GatewayClass
```
nak3, line 94: This will be a same value with https://github.com/projectcontour/contour/blob/e41d550bf3a96fe7c5035f3dc4cfb3f4967329c4/pkg/config/parameters.go#L366-L369 ?
If so, I think that `The string takes the form of "projectcontour.io/<namespace>/contour".` should be mentioned here too.

nak3 (approved):

youngnick (approved):

stevesloka (approved):

PR #422, opened 2021-08-09
Drop GatewayClass/Gateway controllers

sunjayBhatia:
When a Contour is created, the Operator will simply pass along any
GatewayClass configuration to the Contour instances, no longer only
creating a Contour when a GatewayClass/Gateway are present. Removes all
logic in the Operator around Gateway API Resource reconciliation, leaves
it all up to Contour. Also removes all finalizer logic the Operator was
adding to GatewayClasses/Gateways.

Now, in order to use Gateway API, a user will create a Contour config
CRG with the Gateway Controller Name they wish Contour (not the
Operator) to reconcile. The Operator will simply create a Contour
instance and Envoy DaemonSet when this resource is created and Contour
will handle reconciling GatewayClasses/Gateways/*Routes.

Future work to add on to this:
- Remove Gateway API resource RBAC
- Ensure only 1 Contour can be created with a given Gateway Controller
Name
- Move to using Contour config CRD (rather than Operator config CRD) to
configure Contour
- Once Contour takes over managing the Envoy data-plane, remove that
logic from Operator.

Draft for now, based on https://github.com/projectcontour/contour-operator/pull/420 and https://github.com/projectcontour/contour-operator/pull/421 and requires https://github.com/projectcontour/contour/pull/3881 to be merged before this


sunjayBhatia:
Note: tested locally with the Contour image from https://github.com/projectcontour/contour/pull/3881 and it works, this CI will fail until that is merged

api/v1alpha1/contour_types.go
```diff
@@ -91,6 +91,14 @@ type ContourSpec struct {
 	// +optional
 	GatewayClassRef *string `json:"gatewayClassRef,omitempty"`
```
stevesloka, line 92: We can deprecate the `gatewayClassRef` now as well. The `controllerName` is used for determining the gatewayclass.

stevesloka: I need to dig more into this, but one small comment if you want to tackle here or in another PR.

sunjayBhatia: Yeah I wasn’t sure whether we could just remove it now or not bc of API versioning, do we need to do a v1alpha2 to remove it? Or just add a comment/annotation if we can

stevesloka:
@sunjayBhatia looks like the codegen job is failing. I merged the mirror PR in contour so was rerunning tests here to validate.

stevesloka (approved):

sunjayBhatia:
> @sunjayBhatia looks like the codegen job is failing. I merged the mirror PR in contour so was rerunning tests here to validate.

do you want to review all the changes together? I also had separated out the two pre-requisites for this which were a bit big for just this one PR (links in description)

stevesloka:
Ah I merged #421 and there are some conflicts now @sunjayBhatia.

sunjayBhatia:
> Ah I merged #421 and there are some conflicts now @sunjayBhatia.

wdyt about merging the testing one without a second approval, I think its probably less controversial imo

stevesloka:
I'm fine to merge this if you're ok with it @sunjayBhatia. This should fixup a bunch of issues currently.

Issue #419, opened 2021-07-30
test-example-linux job occasionally hangs


skriss:
E.g. https://github.com/projectcontour/contour-operator/runs/3197940834?check_suite_focus=true

Seems to be in the middle of the HTTP request loop.

Can try adding `curl`'s `--max-time` flag to start.

sunjayBhatia:
@skriss when youre back, I attached a PR that should close this, but feel free to take it over if you had something else in mind

skriss:
:+1: LGTM, hopefully that resolves it

PR #424, opened 2021-08-11
Add time limit to example test curl

sunjayBhatia:
Also refactor so both tests actually have to pass not just the second
one

Fixes: #419


stevesloka:
Does this need #422 to work?

sunjayBhatia:
> Does this need #422 to work?

nope should be totally independent

sunjayBhatia:
using powers to merge for CI cleanliness

PR #427, opened 2021-08-12
Remove strange number `|2` in configmap

nak3:
Contour operator creates configmap with the strange `contour.yaml: |2` as below:

```
$ kubectl get cm -n projectcontour contour -o yaml
apiVersion: v1
data:
  contour.yaml: |2

    #
    # server:
    #   determine which XDS Server implementation to utilize in Contour.
    #   xds-server-type: contour
```

Actually `|2` is valid yaml but it makes confused.

This patch fixes it.

### DIFF
--- a/internal/objects/configmap/configmap.go
+++ b/internal/objects/configmap/configmap.go
@@ -38,8 +38,7 @@ const (
 	ContourConfigMapName = "contour"
 )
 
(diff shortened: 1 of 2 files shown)


skriss (approved):
LGTM, can confirm this alleviates the `|2`. Thanks @nak3!

youngnick (approved):

PR #430, opened 2021-08-24
Envoy: add envoy-admin volume, update to 1.19.1

### LINKED ISSUE #423: Contour Deployment/Envoy DaemonSet should be updated for secure Admin listener
When https://github.com/projectcontour/contour/pull/3934 is merged, the Operator will need to match how it sets up the Contour Deployment/Envoy DaemonSet to use the new secure by default Admin Listener on a Unix socket.

skriss:
Adds the envoy-admin emptyDir volume to the Envoy
daemon set and updates the Envoy version to 1.19.1
to match projectcontour/contour.

Closes #423.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -15,7 +15,7 @@ package config
 
 const (
 	DefaultContourImage           = "docker.io/projectcontour/contour:main"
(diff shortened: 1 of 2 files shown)


sunjayBhatia (approved):
only other thing i could think of would be to add some detailed checks of the daemonset to the controller tests, but might be overkill, LGTM

PR #435, opened 2021-08-31
remove BackendPolicy RBAC

skriss:
The upcoming v1alpha2 release of Gateway API drops
BackendPolicy. Contour has dropped references to it
in 7013489dac80e71b2373240a83fe451162f4a4c5. This PR
removes RBAC setup for it from the operator.

### DIFF
--- a/internal/objects/clusterrole/cluster_role.go
+++ b/internal/objects/clusterrole/cluster_role.go
@@ -104,12 +104,12 @@ func desiredClusterRole(name string, contour *operatorv1alpha1.Contour) *rbacv1.
 	gateway := rbacv1.PolicyRule{
 		Verbs:     verbGLWU,
 		APIGroups: groupGateway,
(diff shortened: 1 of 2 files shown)


sunjayBhatia (approved):

skriss:
Looks like the upgrade E2E test does not work well with changes like this, since when deploying the "old" version of the Operator, the old RBAC for the Operator itself isn't applied, so it's using the "new" RBAC with the "old" Operator.

skriss:
We're going to need to fix the upgrade test at some point, but for now I added back in RBAC for the Operator to BackendPolicy, so the tests will continue to pass. This PR now just removes BackendPolicy from the ClusterRole that the Operator creates when instantiating a Contour. After v1.19 ships, can drop the RBAC for the Operator itself.

youngnick (approved):

PR #439, opened 2021-09-01
use GHCR image by default

skriss:
Updates examples, CI and docs to use
the GHCR image by default. The Docker
Hub image will continue to be pushed
for the time being to help migration.

### DIFF
--- a/.github/workflows/build_main.yaml
+++ b/.github/workflows/build_main.yaml
@@ -18,7 +18,7 @@ jobs:
       env:
         DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
         DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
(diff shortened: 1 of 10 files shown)


skriss:
Problems with the upgrade test since it's just manipulating the image tag, rather than applying full manifests from the previous released version. If we manually push some previous releases into GHCR including the `latest` tag then this should be able to pass.

youngnick:
> If we manually push some previous releases into GHCR including the `latest` tag then this should be able to pass.

I think we (well, you :) ) should do this without delay, since it's a one-off, and images not used on Docker Hub will be removed after 6 months.

sunjayBhatia (approved):

skriss:
@stevesloka could you take a quick look at this one when you get a chance? Thanks!

stevesloka (approved):

PR #441, opened 2021-09-03
Fix example test namespace deletion

sunjayBhatia:
use e2e package delete function that stops waiting once namespace is terminating


skriss (approved):
:+1: runs faster now too

sunjayBhatia:
going to force merge to hopefully fix CI issues over the weekend

PR #443, opened 2021-09-07
Fix operator test data race

sunjayBhatia:
Don't shadow namespace key


skriss (approved):
Looks fine, I'm not seeing the race though

stevesloka (approved):

Issue #438, opened 2021-08-31
Fix upgrade test to deal with RBAC or other manifest changes across versions


skriss:
Looks like the upgrade E2E test does not work well with changes like this, since when deploying the "old" version of the Operator, the old RBAC for the Operator itself isn't applied, so it's using the "new" RBAC with the "old" Operator.

_Originally posted by @skriss in https://github.com/projectcontour/contour-operator/issues/435#issuecomment-909390461_

The upgrade test really needs to fully deploy the old version, including all YAML manifests, and then do a full upgrade by applying all of the new manifests, to handle changes like this. Just changing the image tag is not sufficient here.

PR #456, opened 2021-10-16
update to Gateway API v1alpha2

skriss:


skriss:
This is not passing the upgrade test and probably won't without some work - mainly due to https://github.com/projectcontour/contour-operator/issues/438 and a similar issue with the CRD defs themselves (since they've moved groups).

PR #477, opened 2022-01-13
update Gateway API to v0.4.1

skriss:


skriss:
Power-merging to avoid CI complaining since this is a minor dep bump.

PR #478, opened 2022-01-21
examples: move NodePort example to default namespace

skriss:
Puts the Contour CR for the NodePort
examples into the default namespace
to be consistent with other examples.

(This helps with updating the getting started guide, in https://github.com/projectcontour/contour/pull/4288).


skriss:
Merging this in support of https://github.com/projectcontour/contour/pull/4288.

PR #485, opened 2022-02-09
move Config type into internal/operator

skriss:
Drops the internal/config package and moves the
Config struct into internal/operator where it's
used.

### DIFF
--- a/internal/objects/daemonset/daemonset_test.go
+++ b/internal/objects/daemonset/daemonset_test.go
@@ -18,7 +18,6 @@ import (
 	"testing"
 
 	operatorv1alpha1 "github.com/projectcontour/contour-operator/api/v1alpha1"
-	"github.com/projectcontour/contour-operator/internal/config"
 	objcontour "github.com/projectcontour/contour-operator/internal/objects/contour"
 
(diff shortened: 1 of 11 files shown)


internal/objects/daemonset/daemonset_test.go
```diff
@@ -163,8 +162,8 @@ func TestDesiredDaemonSet(t *testing.T) {
 		NetworkType: operatorv1alpha1.LoadBalancerServicePublishingType,
 	}
 	cntr := objcontour.New(cfg)
-	testContourImage := config.DefaultContourImage
-	testEnvoyImage := config.DefaultEnvoyImage
+	testContourImage := "ghcr.io/projectcontour/contour:test"
```
sunjayBhatia, line 165: +1 for this kind of change, using some test data for these rather than constants etc. that may be "real"

sunjayBhatia (approved):

internal/objects/daemonset/daemonset_test.go
skriss: Cool, yeah I had to do this to break the import cycle but there was no good reason to use the actual constants here.

PR #487, opened 2022-02-09
sync Contour ClusterRole rules with Contour's

skriss:


skriss:
Ah yes, forgot that dropping RBAC breaks the upgrade test due to https://github.com/projectcontour/contour-operator/issues/438.

skriss:
> Ah yes, forgot that dropping RBAC breaks the upgrade test due to #438.

Disabled it again for now, will try to find time to work on fixing #438 so we don't have to keep doing this.

sunjayBhatia (approved):

sunjayBhatia:
thanks for including that removal in this 👍🏽

Issue #524, opened 2022-06-13
certgen job name does not get truncated to 63 characters, failing when using long image tags


javsalgar:
**What steps did you take and what happened:**
- Deploy contour-operator using an image with a long tag. Example: `d76d8b5da2ab549cee82cd99fcf8d747dba564863ccc459cae5a91f50bf9c5d3`
- Deploy a contour object

The contour instance will not get deployed because the certgen job will not be created:

```
         1.6551249492126336e+09	ERROR	controller.contour_controller	Reconciler error	{"name": "contour-sample", "namespace": "srdicontou", "error": "failed to ensure job for contour srdicontou/contour-sample: failed to create job srdicontou/contour-certgen-d76d8b5da2ab549cee82cd99fcf8d747dba564863ccc459cae5a91f50bf9c5d3: Job.batch \"contour-certgen-d76d8b5da2ab549cee82cd99fcf8d747dba564863ccc459cae5a91f50bf9c5d3\" is invalid: spec.template.labels: Invalid value: \"contour-certgen-d76d8b5da2ab549cee82cd99fcf8d747dba564863ccc459cae5a91f50bf9c5d3\": must be no more than 63 characters", "errorCauses": [{"error": "failed to ensure job for contour srdicontou/contour-sample: failed to create job srdicontou/contour-certgen-d76d8b5da2ab549cee82cd99fcf8d747dba564863ccc459cae5a91f50bf9c5d3: Job.batch \"contour-certgen-d76d8b5da2ab549cee82cd99fcf8d747dba564863ccc459cae5a91f50bf9c5d3\" is invalid: spec.template.labels: Invalid value: \"contour-certgen-d76d8b5da2ab549cee82cd99fcf8d747dba564863ccc459cae5a91f50bf9c5d3\": must be no more than 63 characters"}]}
```
**What did you expect to happen:**

The job name is truncated to 63 characters so it passes the k8s validations.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Software version:
- Kubernetes version: (use `kubectl version`):
- Kubernetes installer & version:
- Cloud provider or hardware configuration:
- OS (e.g. from `/etc/os-release`):

PR #525, opened 2022-06-13
fix: :bug: Truncate certgen job name to 63 characters

javsalgar:
Signed-off-by: Javier Salmeron Garcia <jsalmeron@vmware.com>

Fixes https://github.com/projectcontour/contour-operator/issues/524

This PR truncates the certgen job name to 63 characters, and adds a validation test to ensure it works as expected.

### DIFF
--- a/internal/objects/job/job.go
+++ b/internal/objects/job/job.go
@@ -43,8 +43,11 @@ func certgenJobName(contourImage string) string {
 	// [TODO] danehans: Remove and use contour.Name + "-certgen" when
 	// https://github.com/projectcontour/contour/issues/2122 is fixed.
 	res := "contour-certgen-" + objutil.TagFromImage(contourImage)
-	// Truncate to avoid issues with long tags and K8s character limit
-	return res[0:63]
+	if len(res) > 63 {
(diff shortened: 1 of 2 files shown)


internal/objects/job/job.go
```diff
@@ -42,7 +42,12 @@ const (
 func certgenJobName(contourImage string) string {
 	// [TODO] danehans: Remove and use contour.Name + "-certgen" when
 	// https://github.com/projectcontour/contour/issues/2122 is fixed.
-	return "contour-certgen-" + objutil.TagFromImage(contourImage)
+	res := "contour-certgen-" + objutil.TagFromImage(contourImage)
+	if len(res) > 63 {
+		// Truncate to avoid issues with long tags and K8s character limit
+		res = res[0:63]
```
skriss, line 48: Another option I've seen used for this is to compute a hash of the full value and use that when needed, but I think this should work fine.

skriss (approved):

javsalgar:
Thanks for the review! How can I fix the codecov issue?

skriss:
> Thanks for the review! How can I fix the codecov issue?

Don't worry about it, we can ignore.

PR #531, opened 2022-08-09
Bump go to 1.19.0

sunjayBhatia:
Also bump golangci-lint to 1.48.0


sunjayBhatia:
merged with powers since its a small change

PR #533, opened 2022-08-17
Fix codespell: Stop correction of complies->compiles

sunjayBhatia:
See https://github.com/codespell-project/codespell/issues/2273

### DIFF
--- a/.codespell.ignorewords
+++ b/.codespell.ignorewords
@@ -1,1 +1,2 @@
 aks
+complies


sunjayBhatia:
Another option is to stop using the `rare` words/corrections dictionary which is enabled by default and contains the `complies->compiles` correction

sunjayBhatia:
will probably need to make a similar change in contour as well

tsaarni (approved):
I'd assume it is better to add to ignore than to drop [rare dictionary](https://github.com/codespell-project/codespell/blob/master/codespell_lib/data/dictionary_rare.txt) altogether. Corrections there look useful, except "complies".

skriss (approved):

PR #565, opened 2022-12-01
bump golangci-lint action to v3

skriss:


sunjayBhatia (approved):
could also bump golangci-lint actual version up to latest, but not super important

PR #573, opened 2022-12-20
sync CRDs from projectcontour/contour

skriss:

### DIFF
--- a/config/crd/contour/01-crds.yaml
+++ b/config/crd/contour/01-crds.yaml
@@ -995,6 +995,26 @@ spec:
                     description: 'Compute Resources required by contour container.
                       Cannot be updated. More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/'
                     properties:
+                      claims:
+                        description: "Claims lists the names of resources, defined
+                          in spec.resourceClaims, that are used by this container.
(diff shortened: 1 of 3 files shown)


skriss:
Hmm, not sure why this is still showing a CRD error when running envtest-based tests, but not when running in kind clusters..

skriss:
In light of #574, I opted to just disable the envtest-based tests rather than mucking around with kubebuilder versions/etc. since the tests are likely pretty redundant with E2E's anyway

sunjayBhatia:
could the issue be a tools version? envtest etc. is probably pretty out of date here and on my dev machine currently

cool with dropping the tests since yeah we're planning on stopping supporting this anyways (and code hasn't changed tbh)

skriss:
> could the issue be a tools version? envtest etc. is probably pretty out of date here and on my dev machine currently
> 
> cool with dropping the tests since yeah we're planning on stopping supporting this anyways (and code hasn't changed tbh)

Yeah my best guess is it's due to an old envtest/associated binaries, but updating to kubebuilder v3 to get newer versions is not super-straightforward and I didn't see the point of going down that road. I think as long as we have the E2E's working we still have decent confidence here.

sunjayBhatia (approved):

